PR #17, opened 2019-05-10
Implement directory support for wf-background

soreau:
Allows users to set the background image key to a directory.

### DIFF
--- a/src/background/background.cpp
+++ b/src/background/background.cpp
@@ -1,6 +1,10 @@
+#include <wordexp.h>
+#include <glibmm/main.h>
+#include <gtkmm/drawingarea.h>
 #include <gtkmm/window.h>
 #include <gtkmm/image.h>
 #include <gdkmm/pixbuf.h>
+#include <gdkmm/general.h>
 #include <gdk/gdkwayland.h>
 #include <config.hpp>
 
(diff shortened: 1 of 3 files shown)


soreau:
https://streamable.com/yeg88

src/background/background.cpp
```diff
@@ -1,3 +1,7 @@
+#include <sys/stat.h>
+#include <wordexp.h>
+#include <dirent.h>
+#include <gtkmm.h>
```
ammen99: I try to include only the necessary files. Are they too much?

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
-    zwf_wm_surface_v1 *wm_surface = NULL;
+    auto pbuf = background->pbuf;
+    auto pbuf2 = background->pbuf2;
+    auto animation = &background->fade_animation;
+    auto alpha = animation->progress();
+    auto images = background->images;
 
-    Gtk::Window window;
-    Gtk::Image image;
+    if (animation->running())
```
ammen99: Braces {} for if/else. No braces are allowed only when both those things happen:
1. No else part
2. Single-line statement

The code isn't updated everywhere I agree, but this is the style that should be used.
Which reminds me I should create a style guide somewhere ...

src/background/background.hpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    WayfireShellApp *app;
+    WayfireOutput *output;
+    zwf_wm_surface_v1 *wm_surface = NULL;
+
+    BackgroundDrawingArea drawing_area;
+    Gtk::Window window;
+    Gtk::Image image;
+    int scale;
+    int cycle_timeout;
```
ammen99: Is there a reason why we can't use `background_cycle_timeout->as_int()` directly?

src/background/background.hpp
```diff
@@ -0,0 +1,53 @@
...
+    bool inhibited = true;
+    uint background[2] = {0};
```
ammen99: Maybe remain this to `active_background_indices`?

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
+    pbuf = create_from_file_safe(path);
+
+    if (!load_next_background(path, background[1]))
+        return false;
+
+    std::cout << "Loaded " << path << std::endl;
+
+    pbuf2 = create_from_file_safe(images[background[1]]);
+
+    pbuf = pbuf->scale_simple(output_width * scale,
```
ammen99: Style issue: multi-line function calls start on the next line with 4 spaces identation, not aligned to the opening (

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
+            continue;
+
+        auto fullpath = std::string(exp.we_wordv[0]) + "/" + file->d_name;
+
+        struct stat next;
+        if (stat(fullpath.c_str(), &next) == 0)
         {
-            std::cerr << "Error: created window was not a wayland surface" << std::endl;
-            std::exit(-1);
+            if (S_ISDIR(next.st_mode))
```
ammen99: Braces for the if/else

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
-            }
-        } catch (...)
+void
+WayfireBackground::reset_background()
+{
+    fade_animation.start(0.0, 0.0);
+    images.clear();
+    change_bg_conn.disconnect();
+    background[0] = background[1] = 0;
+    if (pbuf)
```
ammen99: braces {}

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
+                throw std::exception();
         }
-    }
 
-    void setup_window()
-    {
-        window.set_resizable(false);
-        window.set_decorated(false);
+        scale = window.get_scale_factor();
+        pbuf = pbuf->scale_simple(output_width * scale,
```
ammen99: identation: 4 spaces instead of align with (

src/background/background.cpp
```diff
@@ -9,117 +13,312 @@
...
+        }
 
-        window.property_scale_factor().signal_changed().connect(
-            sigc::mem_fun(this, &WayfireBackground::set_background));
+        fade_animation.start(0.0, 1.0);
+        drawing_area.queue_draw();
+    } catch (...)
+    {
+        window.remove();
+        if (images.size())
```
ammen99: use {}

src/background/background.hpp
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+class WayfireBackground;
+
+class BackgroundDrawingArea : public Gtk::DrawingArea
+{
+    WayfireBackground *background;
+
+    public:
+    BackgroundDrawingArea(WayfireBackground *background);
```
ammen99: I still don't think we should pass the whole `WayfireBackground` to the drawing area. Here is a design which I think should be not-so-hard to implement (correct me if I'm wrong though):

1. Have fade animation managed completely internally by the DrawingArea (so nothing fade-related in WayfireBackground)
2. Have a method like `show_image(Glib::RefPtr<Gdk::PixBuf> next_image)` to the drawing area. What it does is basically tell the drawing area to fade from the current image to the `new_image`.

This way we'd have the actual rendering/fading in the DrawingArea, while Background manages what is to be shown. You could then store `pbuf`, `pbuf2` etc in DrawingArea.

soreau: It took a bit of back and forth but I think I managed to work it out like you described.

src/background/background.cpp
soreau: Yep.

src/background/background.cpp
soreau: Done.

src/background/background.cpp
soreau: Actually this part falls away with your suggestion below.

src/background/background.cpp
soreau: Check.

src/background/background.cpp
soreau: This is how it was before I touched it, will change all instances.

src/background/background.hpp
soreau: That seems a bit long, how about just active_background? Going with this for now.

src/background/background.hpp
soreau: It's a convenience variable but I moved it out of the header and just made them local variables, if that's ok with you.

src/background/background.cpp
soreau: I have been just trying to use braces for both `if` and `else`, when one or the other has more than one line, or none at all when both have a single line, but will use braces from now on unless it's a single `if` followed by single line.

src/background/background.cpp
soreau: If I try to remove gtkmm.h, I get `error: ‘signal_timeout’ is not a member of ‘Glib’` and wordexp.h is also required. I removed sys/stat.h and dirent.h however.

ammen99: The header for `signal_timeout` is `#include <glibmm/main.h>` according to https://developer.gnome.org/glibmm/stable/classGlib_1_1SignalTimeout.html

src/background/background.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+class WayfireBackground;
+
+class BackgroundDrawingArea : public Gtk::DrawingArea
+{
+    wf_duration fade;
+    Glib::RefPtr<Gdk::Pixbuf> pbuf, pbuf2;
```
ammen99: Could we use a bit more descriptive names? Maybe current_image and last_image or something else you find better?

src/background/background.hpp
```diff
@@ -0,0 +1,51 @@
...
+class WayfireBackground
+{
+    WayfireShellApp *app;
+    WayfireOutput *output;
+    zwf_wm_surface_v1 *wm_surface = NULL;
+
+    BackgroundDrawingArea drawing_area;
+    std::vector<std::string> images;
+    Gtk::Window window;
+    Gtk::Image image;
```
ammen99: I feel like this isn't needed anymore ...

src/background/background.cpp
```diff
@@ -9,117 +11,305 @@
...
+    }
+    return true;
+}
+
+bool
+WayfireBackground::load_next_background(Glib::RefPtr<Gdk::Pixbuf> &pbuf, std::string &path)
+{
+    while (!pbuf)
+    {
+        if (++current_background > images.size() - 1)
```
ammen99: one-liner: `current_background = (current_background + 1) % images.size()`

src/background/background.cpp
```diff
@@ -9,117 +11,305 @@
...
+            current_background = 0;
+
+        path = images[current_background];
+        pbuf = create_from_file_safe(path);
+
+        if (!pbuf)
+            images.erase(images.begin() + current_background);
+
+        if (!images.size())
```
ammen99: Shouldn't this be in the very start of the loop body? This way the function also handles empty images vector.

src/background/background.cpp
soreau: That might be, but if I try to remove gtkmm.h, I get a whole slew of errors. Do you want me to figure out exactly what headers are required or why gtkmm.h is necessary to include?

src/background/background.hpp
soreau: Would to_pbuf and from_pbuf work?

src/background/background.hpp
soreau: Removed.

src/background/background.cpp
soreau: Great.

src/background/background.cpp
soreau: Yes, you're right.

src/background/background.cpp
soreau: Actually, now I've whittled it down to `error: ‘Gdk::Cairo’ has not been declared` but can't seem to figure out which header is needed.

soreau: It works if I include glibmm/main.h, gtkmm/drawingarea.h and gdkmm/general.h.

src/background/background.hpp
ammen99: Yep. Maybe it is best to add a comment to describe them better.

soreau:
What is the status of this PR?

ammen99:
I've forgotten about this. There don't seem to be any issues so ...

PR #20, opened 2019-08-21
clock: Fix calender month

soreau:
GDateTime uses month 1-12 while GCalender uses 0-11.

Thanks to gimpnixon for pointing out the problem.


ammen99:
Lol that's true, I also just noticed that it shows September when I open the popup. Thanks!

PR #23, opened 2019-08-27
Add volume widget

soreau:


src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,288 @@
...
+    wf_volume->gvc_stream = gvc_mixer_control_get_default_sink(gvc_control);
+    if (!wf_volume->gvc_stream) {
+        printf("GVC: Failed to get default sink\n");
+        return;
+    }
+
+    if (wf_volume->gvc_channel_map)
+        g_signal_handler_disconnect(wf_volume->gvc_channel_map, wf_volume->volume_changed_g_signal);
+
+    wf_volume->gvc_channel_map = (GvcChannelMap *) gvc_mixer_stream_get_channel_map(wf_volume->gvc_stream);
```
ammen99: sorry I'm giving this information too late, I just found out. Instead of going through the channel map, you can connect directly to the `notify::volume` signal on the gvc_stream

soreau: @ammen99 I would need to see some code or more context to know what you mean.

ammen99: Instead of connecting to the `volume-changed` signal of the channel map, connect to the `notify::volume` signal of the stream itself. That way there is no need to store the channel map too.

ammen99: Forgot to say everything else stays the same basically

soreau: Like I said, I'd have to see some code.

soreau: I looked into what you said and there is no "notify::volume" in gvc. I tried hooking it up to the stream and the function simply wasn't called. The way I found volume-changed was by just reading the headers, and it works.

soreau: If I google `"notify::volume" g_signal_connect` there's only one result and they are passing in a ClutterActor* as the object to g_signal_connect, not a gvc stream.

ammen99:
@soreau, maybe we should also listen for changes to the `is-muted` property, the relevant signal should be `notify::is-muted`

soreau:
@ammen99 I want to see proper documentation on all these signal names first.

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    return OOR;
+}
+
+void
+WayfireVolume::update_icon()
+{
+    volume_level last, current;
+
+    last = get_volume_level(last_volume);
```
ammen99: style issue, `volume_level last = ...`

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    current = get_volume_level(current_volume);
```
ammen99: same, should be `volume_level current = ...`

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+        gvc_mixer_stream_set_volume(gvc_stream, 0.0);
+        main_image.set_from_icon_name("audio-volume-muted", Gtk::ICON_SIZE_MENU);
+    }
+}
+
+bool
+WayfireVolume::on_popover_timeout(int timer)
+{
+    button->get_popover()->hide();
```
ammen99: use `popup()` instead of `hide()`,because it provides animations

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    return false;
+}
+
+void
+WayfireVolume::reset_popover_timeout()
+{
+    if (scale_pressed || volume_clicked)
```
ammen99: Two questions:

1. I get why we don't want to hide when scale is pressed. But in which use-case do we need to check for volume-clicked?
2. Could it be so that it is enough to check the scale state flags? Like whether it is selected or focused (https://developer.gnome.org/gtk3/stable/gtk3-Standard-Enumerations.html#GtkStateFlags)
I am asking this because it could simplify the code by not requiring us to track buttons (plus, the button solution is prone to errors, some GTK details changing, and doesn't work for touch/stylus)

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    if (timeout_was_enabled)
+        popover_timeout = Glib::signal_timeout().connect(sigc::bind(sigc::mem_fun(*this,
+            &WayfireVolume::on_popover_timeout), 0), timeout * 1000);
+    scale_pressed = false;
+}
+
+void
+WayfireVolume::on_popover_button_press(GdkEventButton* event)
+{
+    button->get_popover()->hide();
```
ammen99: `popup()` instead of `hide()`

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    timeout = config_section->get_option("volume_display_timeout", "2.5")->as_double();
+
+    volume_size = config_section->get_option("launcher_size",
+        std::to_string(DEFAULT_ICON_SIZE));
+    volume_size_changed = [=] () { update_icon(); };
+    volume_size->add_updated_handler(&volume_size_changed);
+
+    button = std::unique_ptr<WayfireMenuButton> (new WayfireMenuButton(config));
+    button->add(main_image);
+    auto style = button->get_style_context();
```
ammen99: If this is the part of the code which makes the button not highlighted when the mouses is over it, then I'd say this is inconsistent with other widgets with popovers (menu, clock)

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+    volume_scale.show();
+    hbox.show();
+    main_image.show();
+    button->show();
+}
+
+void
+WayfireVolume::focus_lost()
+{
+    button->get_popover()->hide();
```
ammen99: `popup()` instead of `hide()`

src/panel/widgets/volume.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+void
+WayfireVolume::on_popover_button_press(GdkEventButton* event)
+{
+    button->get_popover()->hide();
+}
+
+static void
+notify_volume (GvcMixerControl *gvc_control,
```
ammen99: For the three callbacks (`notify_volume`, `notify_is_muted`, `default_sink_changes`) it would be more C++-like to have the callbacks call a function in the WayfireVolume class which does the operation, otherwise we have to make a ton of member variables public.

src/panel/widgets/volume.cpp
soreau: I love this style and I want other widgets like menu to use it too. I definitely do not want to see it changed.

src/panel/widgets/volume.cpp
soreau: I'll leave the C++-ifying up to you.

ammen99:
Closing this in favor of merging the `-v2` branch. Thanks soreau!

soreau:
:+1:

PR #25, opened 2019-08-28
menu: Make button icon configurable

### LINKED ISSUE #24: Add option to set menu's icon
Would be nice to have a setting in wf-shell.ini to choose which menu icon you'd like as you're abel to do with the desktop background, right now you can switch the files around and rename them and that works, but if an update is passed your files will be overwritten with the default.

 for example

[menu icon]
.# Full path to image or directory of images
iconimage = /usr/share/icons/gnome/256x256/status/trophy-gold.png

soreau:
Fixes #24.

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -229,21 +229,45 @@ void WayfireMenu::on_popover_shown()
 bool WayfireMenu::update_icon()
 {
     std::string icon;
+    int size = menu_size / LAUNCHERS_ICON_SCALE;
+    Glib::RefPtr<Gdk::Pixbuf> ptr_pbuff;
+    bool error = false;
+
     if (((std::string) menu_icon).empty())
     {
-        icon = ICONDIR "/wayfire.png";
+        icon = default_icon;
     }
     else
     {
         icon = menu_icon;
     }
 
-    int size = menu_size / LAUNCHERS_ICON_SCALE;
-
     button->set_size_request(size, 0);
-    auto ptr_pbuff = Gdk::Pixbuf::create_from_file(menu_icon,
-        size * main_image.get_scale_factor(),
-        size * main_image.get_scale_factor());
+    try
+    {
+        ptr_pbuff = Gdk::Pixbuf::create_from_file(icon,
+            size * main_image.get_scale_factor(),
+            size * main_image.get_scale_factor());
+    }
+    catch(Glib::FileError)
+    {
+        std::cout << "Error loading file: " << icon << std::endl;
+        error = true;
+    }
+    catch(Gdk::PixbufError)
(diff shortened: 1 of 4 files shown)


src/panel/widgets/menu.hpp
```diff
@@ -70,8 +70,10 @@ class WayfireMenu : public WayfireWidget
     void update_popover_layout();
 
     wf_option fuzzy_search_enabled;
-    wf_option menu_size;
+    wf_option menu_icon, menu_size;
+    wf_option_callback menu_icon_changed;
     wf_option_callback menu_size_changed;
+    std::string default_icon = ICONDIR "/wayfire.png";
```
ammen99: why does this need to be declared in the header?

src/panel/widgets/menu.cpp
```diff
@@ -231,11 +231,35 @@ void WayfireMenu::on_popover_shown()
...
 {
     int size = menu_size->as_int() / LAUNCHERS_ICON_SCALE;
+    Glib::RefPtr<Gdk::Pixbuf> ptr_pbuff;
+    bool error = false;
 
     button->set_size_request(size, 0);
-    auto ptr_pbuff = Gdk::Pixbuf::create_from_file(ICONDIR "/wayfire.png",
-        size * main_image.get_scale_factor(),
-        size * main_image.get_scale_factor());
+    try
```
ammen99: Instead of a try-catch, we can use https://github.com/WayfireWM/wf-shell/blob/master/src/util/gtk-utils.hpp#L8

ammen99: Also, if you would like to have error messages, I'm not opposed at all. Add them to the functiton mentioned above (currently it just returns null without further notice)

src/panel/widgets/menu.hpp
ammen99: I still think this should be moved into the source file.

ammen99:
Please move the code into the source file as in the comment. Otherwise ready for merge, finally :)

src/panel/widgets/menu.hpp
```diff
@@ -65,11 +65,12 @@ class WayfireMenu : public WayfireWidget
...
 
-    WfOption<std::string> panel_position{"panel/position"};
-    void update_popover_layout();
-
     WfOption<bool> fuzzy_search_enabled{"panel/menu_fuzzy_search"};
+    WfOption<std::string> panel_position{"panel/position"};
     WfOption<int> menu_size{"panel/launcher_size"};
+    WfOption<std::string> menu_icon{"panel/menu_icon"};
+    void update_popover_layout();
+    std::string default_icon;
```
ammen99: Why does this need to be in the class at all :P

PR #27, opened 2019-08-31
background: Add color option

soreau:
This allows for setting color in the [background] section in the       
format '\\#XXXXXX' where X is a hexidecimal digit. The hashtag
must be escaped with '\\' since it's also used for comments. Fixes 26.

### DIFF
--- a/src/background/background.cpp
+++ b/src/background/background.cpp
@@ -5,6 +5,7 @@
 #include <gtkmm/image.h>
 #include <gdkmm/pixbuf.h>
 #include <gdkmm/general.h>
+#include <gtkmm/cssprovider.h>
 #include <gdk/gdkwayland.h>
 #include <config.hpp>
 
@@ -17,6 +18,8 @@
 
 #include "background.hpp"
 
+
+#define DEFAULT_BACKGROUND_COLOR "#333333"
 
 void
 BackgroundDrawingArea::show_image(Glib::RefPtr<Gdk::Pixbuf> image, double offset_x, double offset_y)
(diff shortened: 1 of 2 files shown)


src/background/background.cpp
```diff
@@ -289,6 +292,44 @@ WayfireBackground::reset_cycle_timeout()
...
+validate_color_string(std::string c)
+{
+    int i;
+
+    if (c.size() != 7 || c[0] != '#')
+        return false;
+
+    for (i = 1; i < 7; i++)
+    {
+        if (!isxdigit(c[i]))
```
ammen99: put an `std::`

src/background/background.cpp
```diff
@@ -289,6 +292,44 @@ WayfireBackground::reset_cycle_timeout()
...
+    auto color = background_color->as_string();
+    auto style_context = window.get_style_context();
+
+    if (!validate_color_string(color))
+    {
+        std::cerr << "Ignoring invalid color: \"" << color << "\"" << std::endl;
+        std::cerr << "Color must be in format \\#XXXXXX where X is a hexidecimal digit." << std::endl;
+        color = DEFAULT_BACKGROUND_COLOR;
+    }
+std::cout << "Setting color: " << color << std::endl;
```
ammen99: identation

src/background/background.cpp
soreau: Done.

src/background/background.cpp
soreau: This was a debug message. Removed.

ammen99:
Closing since this has been implemented in Wayfire itself.

PR #39, opened 2019-11-17
Make pulse dependency optional

soreau:
This makes building gvc and volume widget optional which is needed
to build wf-shell without pulseaudio dependency.

### DIFF
--- a/src/panel/panel.cpp
+++ b/src/panel/panel.cpp
@@ -12,6 +12,7 @@
 #include <sstream>
 
 #include <map>
+#include "../../config.h"
 
 #include "panel.hpp"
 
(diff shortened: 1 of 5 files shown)


soreau:
Yea, I think I can do it with a single line like [this](https://github.com/WayfireWM/wf-shell/blob/master/meson.build#L30). I can't recall a specific reason for adding a config.h file.

src/panel/panel.cpp
```diff
@@ -201,8 +203,10 @@ class WayfirePanel::impl
             return Widget(new WayfireNetworkInfo());
         if (name == "battery")
             return Widget(new WayfireBatteryInfo());
+#ifdef HAVE_PULSE
         if (name == "volume")
```
ammen99: I'd also add an `#else` to print a warning if volume was disabled (otherwise it might be confusing for the user)

soreau: Added.

ammen99: I meant runtime warning if the volume widget is specified in the config file :)

ammen99:
LGTM.

PR #43, opened 2019-12-23
build: Add xml files to the repository

soreau:


meson.build
```diff
@@ -28,6 +28,7 @@ add_project_arguments(['-Wno-pedantic', '-Wno-unused-parameter', '-Wno-parenthes
 
 icon_dir = join_paths(get_option('prefix'), 'share', 'wayfire', 'icons')
 add_project_arguments('-DICONDIR="' + icon_dir + '"', language : 'cpp')
+add_project_arguments('-DPLUGIN_XML_DIR="' + join_paths(get_option('prefix'), 'share/wayfire/metadata') + '"', language : 'cpp')
```
ammen99: Hm, the only thing I don't really like is that it may turn out that wayfire and wf-shell's XMLs are installed at different location (since the wf-shell XMLs should be part of this repo). But maybe this is not our fault then ...

soreau: We either need to 1) make a rule that there is one directory used relative to wayfire install prefix or 2) make some way to specify multiple directories to look. I vote for 1) because I think it is the simple solution. I was thinking you'd want to s/metadata/xml/.

soreau: My idea is that wf-shell and friends can depend on wayfire.pc so it can find where to look.

PR #44, opened 2019-12-27
background: render background with scale

ammen99:


src/background/background.cpp
```diff
@@ -22,16 +22,18 @@ void BackgroundDrawingArea::show_image(Glib::RefPtr<Gdk::Pixbuf> image,
 {
     if (!image)
     {
-        to_image.pbuf.clear();
-        from_image.pbuf.clear();
+        to_image.source.clear();
+        to_image.source.clear();
```
soreau: Is there a reason you're calling this twice in a row? Perhaps one is meant to be `to_image`?

ammen99: oops

PR #54, opened 2020-05-02
src/meson.build: generate wf-shell.pc

alyssais:
This is necessary to allow wcm to look up wf-shell's metadata and
sysconf directories, to initialize wf-shell's configuration.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -3,3 +3,13 @@ subdir('panel')
 subdir('background')
 subdir('dock')
 
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(
+  version: meson.project_version(),
+  name: meson.project_name(),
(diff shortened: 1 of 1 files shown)


src/meson.build
```diff
@@ -3,3 +3,13 @@ subdir('panel')
 subdir('background')
 subdir('dock')
 
+pkgconfig = import('pkgconfig')
+pkgconfig.generate(
+  version: meson.project_version(),
+  name: meson.project_name(),
+  description: 'GTK3-based panel for Wayfire',
```
soreau, line 10: As indicated by the calls to subdir above, wf-shell is a `GTK3 shell for Wayfire` since it includes panel, dock and background.

alyssais: Ah, yes. Fixed. :)

PR #58, opened 2020-08-14
Add "Fill and Crop" option for wf-background

CheesecakeCG:
This is my first time ever making a PR, sorry in advance if I messed something up!

I added an option that makes the wallpaper fit the height of the screen and crop the sides to preserve the aspect ratio. 
I haven't tested this on anything else besides a single 5:4 monitor with 16:9 and 21:9 wallpapers, but it works in that case. 

[Demo Video](https://streamable.com/0iwjmm)

I also realized while writing this that it doesn't crop vertical wallpapers correctly, making the name "Fill and Crop" somewhat misleading. I'll fix that ASAP.

### DIFF
--- a/src/background/background.cpp
+++ b/src/background/background.cpp
@@ -72,34 +72,73 @@ WayfireBackground::create_from_file_safe(std::string path)
     int width = window.get_allocated_width() * scale;
     int height = window.get_allocated_height() * scale;
 
+    float screen_aspect_ratio = (float) width / height;
+    float image_aspect_ratio;
+
+    std::string fill_and_crop_string = "fill_and_crop";
+    std::string preserve_aspect_string = "preserve_aspect";
+    std::string stretch_string = "stretch";
+
+	enum FillMode { fill_and_crop, preserve_aspect, stretch };
+	FillMode current_mode;
+	
+    if (!fill_and_crop_string.compare(background_fill_mode))
+    {        
+    	current_mode = fill_and_crop;
+    }
+    else
+    {
+    	if (!preserve_aspect_string.compare(background_fill_mode))
+    	{
+    	    current_mode = preserve_aspect;
+    	} 
+    	else
+    	{
+            current_mode = stretch;
+        }
+    }
+    
     try {
         pbuf =
-            Gdk::Pixbuf::create_from_file(path, 
-                background_fill_and_crop ? -1 : width, height,
-                background_fill_and_crop || background_preserve_aspect);
+            Gdk::Pixbuf::create_from_file(path, width, height,
+                current_mode != stretch);
     } catch (...) {
         return Glib::RefPtr<Gdk::Pixbuf>();
     }
 
-    if (background_fill_and_crop)
-    {        
-     	offset_x = (width - pbuf->get_width()) * 0.5;
-        offset_y = (height - pbuf->get_height()) * 0.5;
+    if (current_mode == fill_and_crop) {
+	    image_aspect_ratio = (float) pbuf->get_height() / pbuf->get_width();
+		bool should_fill_width = 
+		     !((image_aspect_ratio < 1) && (screen_aspect_ratio > 1));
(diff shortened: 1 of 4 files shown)


src/background/background.cpp
```diff
@@ -72,34 +72,73 @@ WayfireBackground::create_from_file_safe(std::string path)
...
-            offset_x = eq_width ? 0 : (width - pbuf->get_width()) * 0.5;
-            offset_y = eq_width ? (height - pbuf->get_height()) * 0.5 : 0;
+    	    current_mode = preserve_aspect;
     	} 
     	else
     	{
-        offset_x = offset_y = 0.0;
+            current_mode = stretch;
         }
     }
```
CheesecakeCG: I'm sure there's a better way to check what the available options are, but I could not find it.

ammen99: Not really, wf-config doesn't support enum types yet (if I understand you correctly)

CheesecakeCG: Yeah, that's what I meant! Thanks!

soreau:
Does your fill-and-crop mode work for images of random sizes, i.e. not matching width or height of the output? Here, when I try a randomly sized images, the mode doesn't work very well. Typically, the bottom half of the image is cut off and the top isn't at all. Also, when I select other modes that set offset as a result of image not being same dimensions as output, and then switch back to fill-and-crop, the offset isn't reset, so then all of the images are offset by whatever the last image was for the other mode.

CheesecakeCG:
> then all of the images are offset by whatever the last image was for the other mode.
I'd accidentally put an "else" in right before it would have set the offset for new image. Oops.

> Does your fill-and-crop mode work for images of random sizes, i.e. not matching width or height of the output?
Yes, it does now at least. It works on vertical screens too.

wf-shell.ini.example
```diff
@@ -3,8 +3,11 @@
 [background]
 # Full path to image or directory of images
 # image = /usr/share/wayfire/wallpaper.jpg
-# Whether to scale images or preserve background ratio
+# Whether to scale images or preserve background ratio and fit width
```
ammen99, line 6: As far as I understand, the `preserve_aspect` option is no longer valid, so you should remove it from the example config file.

src/background/background.cpp
```diff
@@ -72,25 +72,72 @@ WayfireBackground::create_from_file_safe(std::string path)
     int width = window.get_allocated_width() * scale;
     int height = window.get_allocated_height() * scale;
 
+    float screen_aspect_ratio = (float) width / height;
```
ammen99, line 75: Indentation seems to be quite broken on a lot of places. Use spaces only.

ammen99: And add 4 spaces for each new indentation level.

src/background/background.cpp
```diff
@@ -72,25 +72,72 @@ WayfireBackground::create_from_file_safe(std::string path)
...
+    float image_aspect_ratio;
+
+    std::string fill_and_crop_string = "fill_and_crop";
+    std::string preserve_aspect_string = "preserve_aspect";
+    std::string stretch_string = "stretch";
+
+	enum FillMode { fill_and_crop, preserve_aspect, stretch };
+	FillMode current_mode;
+	
+    if (!fill_and_crop_string.compare(background_fill_mode))
```
ammen99, line 85: I would rather do these `if/else if/else` guards one time, and potentially have `gdk::create_from_file` called in each of these branches with the correct parameters, special casing here doesn't look very good.

ammen99 (changes requested):
Sorry for the delay, I did not get a notification when you requested my review.

See the individual comments.

botiapa:
Sorry for the bump, but any progress on this? IMHO this is quite an important feature, and I'd very much like to see it merged if ready.

ammen99:
Closing in favor of #226

PR #66, opened 2020-09-19
Rename launcher_size to launchers_size

### LINKED ISSUE #65: Rename launcher_size to launchers_size
Then it will be consistent with `launchers_spacing`. It will also cause the warning about the `42` desktop file not existing to go away.

wb9688:
Fixes #65


ammen99:
Thanks!

PR #69, opened 2020-10-20
Update to latest foreign toplevel protocol

soreau:


ammen99:
Do you plan on adding any functionality depending on these new versions? Otherwise there doesn't seem to be a need to update the version.

ammen99:
I suppose for example we could hide the toplevels which have their parent set.

soreau:
> I suppose for example we could hide the toplevels which have their parent set.

I implemented this in the latest push.

ammen99:
Closing this, I suppose we can just merge #70 because it is basically this PR + fixes

PR #70, opened 2020-11-04
Fix crash

soreau:


ammen99 (approved):
LGTM, thanks!

src/panel/widgets/window-list/toplevel.cpp
```diff
@@ -477,9 +523,32 @@ static void handle_toplevel_closed(void *data, toplevel_t handle)
...
+        children.erase(std::find(children.begin(), children.end(), handle));
+    }
     impl->window_list->handle_toplevel_closed(handle);
 }
 
+static void handle_toplevel_parent(void *data, toplevel_t handle, toplevel_t parent)
+{
+    auto impl = static_cast<WayfireToplevel::impl*> (data);
+    impl->set_parent(parent);
+    if (!parent)
```
ammen99, line 539: Here, if the toplevel already had a parent, you should erase the toplevel from the parent's children list.

ammen99: Also, we  you should call handle_output_enter *only* if there was already a parent.

PR #72, opened 2020-11-10
build: Make wayland-logout installation optional

soreau:
If a packager has packaged wayland-logout separately, the
wayland-logout subproject will cause wf-shell to conflict.


epsilon-0:
this looks good :smiley_cat:

ammen99 (approved):
Thanks!

PR #76, opened 2020-12-16
reload menu items when a new .desktop entry is created

### LINKED ISSUE #34: menu does not update when apps are added or removed
only when wf-panel is restarted does the menu update.
it should update with app changes.

hcsubser:
Instead of having to restart wf-panel to show new menu entries, now it is done every time menu is opened.
This should make much better user experience since now i do not have to log out/ kill wf-panel every time i  install a new app.
Also closes #34

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -224,6 +224,11 @@ bool WayfireMenu::on_sort(Gtk::FlowBoxChild* a, Gtk::FlowBoxChild* b)
 void WayfireMenu::on_popover_shown()
 {
     flowbox.unselect_all();
+    for (auto* child : flowbox.get_children())
+ 		{flowbox.remove(*child);}
+	loaded_apps.clear();
+	load_menu_items_all();
+	update_popover_layout();
 }
 
 bool WayfireMenu::update_icon()
@@ -337,9 +342,6 @@ void WayfireMenu::init(Gtk::HBox *container)
     hbox_bottom.pack_end(logout_button, false, false);
     popover_layout_box.pack_end(hbox_bottom);
     popover_layout_box.pack_end(separator);
-
-    load_menu_items_all();
-    update_popover_layout();
 
     hbox.show();
     main_image.show();
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -225,10 +225,10 @@ void WayfireMenu::on_popover_shown()
 {
     flowbox.unselect_all();
     for (auto* child : flowbox.get_children())
- 		{flowbox.remove(*child);}
-	loaded_apps.clear();
-	load_menu_items_all();
-	update_popover_layout();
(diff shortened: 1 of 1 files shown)


ammen99:
Thanks for the patch! However, I am unsure about the performance of this approach. Is the Gio AppInfo database cached somehow? Or does this cause all .desktop files to be read from the disk when opening a menu? Also, take into account that we also read entries from the desktop, which aren't cached at all. In case of many entries in the `~/Desktop` folder this might cause a significant delay.

wb9688:
Can't you use inotify to monitor for file changes and then reload it?

Otherwise there's also [LXDE's libmenu-cache](https://github.com/lxde/menu-cache), but that is yet another dependency (that also comes with its own daemon and stuff) and is more intended for having an actual menu with categories and stuff, though you could also call `menu_cache_list_all_apps` to just get a list of all apps.

hcsubser:
Yes, it currently does read from drive every time. With 43 .desktop files in total in /usr/share/applications and ~/.local/share/applications there is a little bit of delay when opening the menu however it is less then a second and that is on a good old hard drive, I'm pretty sure that on an SSD it would be pretty much instant. Also it seems that Gio AppInfo does cache entries since I notice this slowdown only when a change happens (or on a cold boot).

That being said I did not realize that it also reads files from ~/Desktop, since I never put anything there. Because of that I will be using wf-panel with this patch for some time to see how it performs( on an HDD ofc) and will also push it by making a lot of desktop files just to test if and how much it slows down. I can live with current less than a second time. Still, this menu is still much faster than Pantheon's menu on first initialization( that one is so slow).
After I finish those tests and after using it for some time I will report here the results and based on those results we can decide if this PR can be merged or if we need to implement some caching/inotify.
'Till then I will add "(WIP)" to this PR.

AdrianVovk:
My 2¢:

If you were to use GIO instead of doing your own app info processing, you'd get this for free.

There's an API called GAppInfoMonitor, which sends a signal whenever an app is updated. GIO caches app infos internally, but not the list. So the best course of action is this:

On startup, load the app list. Connect to the monitor. Whenever the monitor fires its event, set a flag that a reload is needed. Then, either the next time the menu opens or whenever gtk idles (using glib's Idle.add), reload the list of apps (usually the monitor fires many events in a row, for example during system updates).

Of course, doing it this way is impossible unless you drop your own search path and use the standard, because GIO only monitors and caches the standard paths.

ammen99:
> Of course, doing it this way is impossible unless you drop your own search path and use the standard

We do use Gio for that, it is simply that we load a few more additional paths (`~/Desktop`). Anyway, the API you suggested should be a good way forward, we should simply reload apps when we get that event or when `~/Desktop` changes, which we could use inotify for.

AdrianVovk:
@ammen99 Ah last I remember you manually looked through /usr/share, ~/.local/share, etc. If you've changed that, then definitely it's much more doable!

PR #77, opened 2020-12-23
Add {panel,dock}/css_path and custom CSS support

diamondburned:
This pull request adds `panel/css_path` and `dock/css_path` into the configuration file schema, allowing for custom CSS loading of both the panel and the dock.

```ini
[panel]
css_path = /path/to.css

[dock]
css_path = /path/to.css
```

I think the code to load those CSS could've been better placed elsewhere, but I really can't be bothered to set up my editor environment again, so I would kindly ask you the maintainers to do that, as you know the project more than I do.

### DIFF
--- a/src/dock/dock.cpp
+++ b/src/dock/dock.cpp
@@ -11,6 +11,7 @@
 #include <wf-autohide-window.hpp>
 
 #include "dock.hpp"
+#include "../util/gtk-utils.hpp"
 
 class WfDock::impl
 {
@@ -19,6 +20,8 @@ class WfDock::impl
     wl_surface *_wl_surface;
 
     Gtk::HBox box;
+
+    WfOption<std::string> css_path{"dock/css_path"};
 
     public:
     impl(WayfireOutput *output)
@@ -34,6 +37,19 @@ class WfDock::impl
         window->signal_size_allocate().connect_notify(
             sigc::mem_fun(this, &WfDock::impl::on_allocation));
         window->add(box);
+
+        if ((std::string)css_path != "")
+        {
+            auto css = load_css_from_path(css_path);
+            if (css)
+            {
+                auto screen = Gdk::Screen::get_default();
+                auto style_context = Gtk::StyleContext::create();
+                style_context->add_provider_for_screen(
+                    screen, css, GTK_STYLE_PROVIDER_PRIORITY_USER);
+            }
+        }
+
         window->show_all();
         _wl_surface = gdk_wayland_window_get_wl_surface(
             window->get_window()->gobj());
--- a/metadata/dock.xml
+++ b/metadata/dock.xml
@@ -3,6 +3,10 @@
 	<plugin name="dock">
 	<_short>Dock</_short>
 	<category>Shell</category>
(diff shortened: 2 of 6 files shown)


soreau:
You are the first one I know of that has expressed a desire to adjust the theme separately from the global gtk scheme. What's wrong with styling the way you'd typically style any other gtk application?

diamondburned:
> What's wrong with styling the way you'd typically style any other gtk application?

I just don't want my bar's button styles to look like a regular application's. It just doesn't look right. Also, having custom CSS lets me toy with some weird styles that I wouldn't otherwise want in a regular application.

On a tangent, the `headerbar` background color doesn't actually look right to me precisely because I theme other applications like that, and I really want to experiment with `background-image`, so that's another reason.

ammen99:
@diamondburned Have you thought about simply applying a custom theme via `GTK_THEME`?

soreau:
@ammen99 Recently I inquired about GTK_THEME in #gtk and one of the developers indicated that GTK_THEME is for testing and meant not to work.

ammen99:
> @ammen99 Recently I inquired about GTK_THEME in #gtk and one of the developers indicated that GTK_THEME is for testing and meant not to work.

Ah, I see. I'll review this PR soon then, it would make sense to have it.

diamondburned:
In hindsight, using the `USER` priority for the CSS provider isn't the best idea, as it made it harder to test CSS using `GTK_DEBUG=interactive`, since the debugger's priority is lower.

soreau:
@diamondburned There are other priorities, you can find the gtk3 list [here](https://developer.gnome.org/gtk3/stable/GtkStyleProvider.html#GTK-STYLE-PROVIDER-PRIORITY-FALLBACK:CAPS), though I'm not sure if there is one that will do what you want. Did you still want someone else to pick this up and submit a new PR? It looks like I might have to do the same for [wf-menu](https://github.com/soreau/wf-extra-protos).

diamondburned:
> @diamondburned There are other priorities, you can find the gtk3 list [here](https://developer.gnome.org/gtk3/stable/GtkStyleProvider.html#GTK-STYLE-PROVIDER-PRIORITY-FALLBACK:CAPS), though I'm not sure if there is one that will do what you want. Did you still want someone else to pick this up and submit a new PR? It looks like I might have to do the same for [wf-menu](https://github.com/soreau/wf-extra-protos).

I think `USER` is fine; it's just a minor inconvenience, and `USER` is technically correct (though `APPLICATION` isn't so bad either).

src/dock/dock.cpp
```diff
@@ -34,6 +37,17 @@ class WfDock::impl
         window->signal_size_allocate().connect_notify(
             sigc::mem_fun(this, &WfDock::impl::on_allocation));
         window->add(box);
+
+        if ((std::string)css_path != "") {
```
ammen99, line 41: style: { should be on a new line. This applies to all `ifs` in this PR

ammen99 (commented):
LGTM overall, please fix code style.

ammen99 (approved):
LGTM, thanks!

PR #81, opened 2021-01-29
Added custom onclick command support to network widget.

jgmdev:
This option allows a user to define a custom onclick command for the network widget so commands like `terminal -e nmtui` or others can be used in place of having a static dependency on gnome-control-center.


ammen99 (approved):
LGTM, thanks!

PR #82, opened 2021-01-31
[wf-dock] improved icon search.

jgmdev:
Fixes application icons not found like Quassel Client where the icon is not named org.kde.quassel but quassel or Wayfire config manager which .desktop file is not named org.gtk.wayfire-config-manager.desktop but instead wayfire-config-manager.desktop.


ammen99 (commented):
Hmm, I think we should finally put this code in some shared place, because it is duplicated for panel's window-list too https://github.com/WayfireWM/wf-shell/blob/master/src/panel/widgets/window-list/toplevel.cpp#L643

Would you like to do that? (there are already shared things in `src/util/`)

jgmdev:
> Would you like to do that? (there are already shared things in `src/util/`)

I think I can do it, maybe it can be moved to src/util/gtk-utils.hpp or create a separate src/util/gtk-iconprovider.hpp file. What do you think is more desirable?

ammen99:
> I think I can do it, maybe it can be moved to src/util/gtk-utils.hpp or create a separate src/util/gtk-iconprovider.hpp file. What do you think is more desirable?

I think it is fine to put it in gtk-utils

src/util/gtk-utils.cpp
```diff
@@ -86,17 +88,208 @@ void set_image_icon(Gtk::Image& image, std::string icon_name, int size,
...
+            }
+        }
+    }
+
+    // Perform a search and select best match
+    if (!app_info)
+    {
+        std::vector<std::string> app_id_variations;
+
+        if (app_id == "Code")
```
ammen99, line 184: I definitely don't want app-specific workarounds here. You can try to add custom icon mappings for specific apps, both panel and dock support it.

src/util/gtk-utils.cpp
```diff
@@ -86,17 +88,208 @@ void set_image_icon(Gtk::Image& image, std::string icon_name, int size,
...
+            app_id_variations.push_back("visual-studio-code");
+
+        app_id_variations.push_back(app_id);
+
+        // If appid has dashes add first component to search
```
ammen99, line 189: This sounds too hacky. Which clients require this?

src/util/gtk-utils.cpp
```diff
@@ -86,17 +88,208 @@ void set_image_icon(Gtk::Image& image, std::string icon_name, int size,
...
+            if(desktop_file != "")
+            {
+                app_info = Gio::DesktopAppInfo::create(desktop_file);
+                break;
+            }
+        }
+    }
+
+    // If app has dots try each component
+    if (!app_info && app_id.find('.') != std::string::npos)
```
ammen99, line 226: I thought we already had a heuristic above (line 143) which handles this case?

ammen99 (changes requested):

src/util/gtk-utils.cpp
wb9688: And it's not even needed, because VS Code correctly set its WMClass to the StartupWMClass defined in its desktop file.

jgmdev: On my arm64 system /usr/share/applications/visual-studio-code.desktop
```
[Desktop Entry]
Name=Visual Studio Code
Comment=Code Editing. Refined.
GenericName=Text Editor
Exec=/opt/visual-studio-code/code --no-sandbox --unity-launch %F
Icon=visual-studio-code
Type=Application
StartupNotify=false
StartupWMClass=code
Categories=Utility;TextEditor;Development;IDE;
MimeType=text/plain;inode/directory;
Actions=new-empty-window;
Keywords=vscode;

[Desktop Action new-empty-window]
Name=New Empty Window
Exec=/opt/visual-studio-code/code --no-sandbox --new-window %F
Icon=visual-studio-code
```
Maybe since I still use wayfire 0.6 my system is picking up 'Code' as appid not sure if newer wayfire now takes into consideration StartupWMClass when an application is launched which in this case should be 'code', still it wouldn't find the correct icon because the app desktop file is named visual-studio-code and I have codelite and codeblocks installed so in this case that g_desktop_app_info_search() is triggered it would pick any of those for this scenario. 

This is why on IRC I suggested the crazy idea of taking into consideration the app title when searching with appid fails, since many applications title can be "some file - App Name" or "App Name - some file" or "App Name" which I have noticed in most cases matches with app-name.desktop file.

Will remove this silly workaround I did out of desperation :) even the crazy app title lookup would have been better than this.

jgmdev: >  I have codelite and codeblocks installed so in this case that g_desktop_app_info_search() is triggered it would pick any of those for this scenario.

Unless I open each desktop file returned by g_desktop_app_info_search() and verify the best matching StartupWMClass

src/util/gtk-utils.cpp
jgmdev: One example would be blueman which has .desktop files:
* blueman-adapters.desktop  
* blueman-manager.desktop
* bluetooth-sendto.desktop

If I click on the blueman tray icon (using waybar for tray support) and select 'Local Services' the reported appid is blueman-applet and no icon is found. With the discussed code an icon related to blueman is properly picked instead of unknown.

src/util/gtk-utils.cpp
jgmdev: On line 143 the search is for the last component of the application, for example, some applications have the appid properly named org.somename.appname but the .desktop file is named appname.desktop so this works on this cases.

Line 226 is when the app id is reported as appname.ext where ext (like in some games) can be the architecture of the application eg: gamename.x86_64 or gamename.aarch64, etc... In this kind of cases a desktop file gamename.desktop exists and with this method it is properly found. We perform this method as last because all previous searching method should have higher priority and for applications that provide a proper appid  (org.name.appname) and desktop file (org.name.appname.desktop) it would result on picking an incorrect icon.

src/util/gtk-utils.cpp
jgmdev: Ok, last commit fixes the issue, sorry for been slow on what you guys meant with the StartupWMClass :)

PR #84, opened 2021-02-28
Add XDG_CONFIG_HOME support

yemouu:
Signed-off-by: yemou <meeyou@tuta.io>

### DIFF
--- a/src/util/wf-shell-app.cpp
+++ b/src/util/wf-shell-app.cpp
@@ -10,9 +10,18 @@
 
 std::string WayfireShellApp::get_config_file()
 {
-    std::string home_dir = getenv("HOME");
-    std::string config_file = home_dir + "/.config/wf-shell.ini";
-    return config_file;
+    std::string config_dir;
+
+    char* config_home = getenv("XDG_CONFIG_HOME");
+    if (config_home == NULL)
+    {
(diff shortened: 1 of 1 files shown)


src/util/wf-shell-app.cpp
```diff
@@ -10,9 +10,13 @@
 
 std::string WayfireShellApp::get_config_file()
 {
-    std::string home_dir = getenv("HOME");
-    std::string config_file = home_dir + "/.config/wf-shell.ini";
-    return config_file;
+    std::string config_dir = getenv("XDG_CONFIG_HOME");
```
ammen99, line 13: The old code has this bug as well, but we'd better fix it while we're at it: getenv returns `NULL` not an empty string when the environment variable is not found. So we cannot assign the result directly to a `std::string`, we first need to store the result in a `char*`, compare with null, and then decide.

yemouu:
Will this work? I think it would be safe to assume that HOME is set.

ammen99 (approved):
LGTM, thanks!

ammen99:
> Will this work? I think it would be safe to assume that HOME is set.

Yes, indeed, HOME is a rather safe assumption .. Ofc we'll crash without it set but I can't imagine why the user wouldn't have this.

Issue #22, opened 2019-08-22
launchers


gimpcheesestick:
Unable to add launcher buttons to more than one widget side. Have a launchers1 and launchers2 so you adding seperate launchers to widget_left, widget_right or widgets_center is functional. Functional and practical
![launchers](https://user-images.githubusercontent.com/54387115/63485099-1dfcd180-c470-11e9-82de-898fddb8c044.png)

librewish:
yes i also want this functionality.
i want to add launchers at left side of panel 
and also at right side of panel.

PR #86, opened 2021-03-14
Add  multi-purpose customizable widgets fo wf-panel

pacomod:
Hi,
This PR adds multi-purpose customizable widgets to wf-panel:
- multi-purpose: music player, backlight (see configuration examples below), and hopefully other things I haven't think of yet…
- customizable: icon, default or dynamic tooltip, mouse buttons and scroll actions can be set by the user.

These are merely 'super'-launchers with a basic widget look (button-like, no animation) and dedicated commands for mouse button/srolll actions and tooltip.
I made this to fit my needs but I hope it can be useful to other wf-shell users as well; it may also be a workaround to issue #22.

**Configuration** (in wf-shell.ini):
`customizables_size = <int>`
`customizables_spacing = <int>`
`customizables_invert = <bool>`
`custom_label_<name> = <string>`  _(mandatory)_
`custom_icon_<name> = <string>`
`custom_btn[1-3]_cmd_<name> = <string>`
`custom_scr_[up|dn]_cmd_<name> = <string>`
`custom_tooltip_cmd_<name> = <string>`
- customizables_* options impact all customizables widgets
- custom_*\_\<name> options are related to custom_label_\<name> widget

**Important note about custom_tooltip_cmd:**
This spawns a synchronous command, therefore waits for it to exit; so be careful with this command.

**Configuration example:**
`customizables_size = 16`
`customizables_spacing = 2`
`customizables_invert = true`

`custom_label_1 = mpc`
`custom_icon_1 = /usr/share/icons/Adwaita/16x16/emblems/emblem-music-symbolic.symbolic.png`
`custom_btn1_cmd_1 = mpc toggle`
`custom_btn2_cmd_1 = mpc consume`
`custom_btn3_cmd_1 = sonata`
`custom_scr_up_cmd_1 = mpc seek -00:00:10`
`custom_scr_dn_cmd_1 = mpc seek +00:00:10`
`custom_tooltip_cmd_1 = mpc current`

`custom_label_2 = backlight`
`custom_icon_2 = /usr/share/icons/Adwaita/16x16/devices/video-display-symbolic.symbolic.png`
`custom_btn1_cmd_2 = brightnessctl --device=intel_backlight set 1`
`custom_btn2_cmd_2 = killall kanshi || /usr/bin/kanshi`
`custom_btn3_cmd_2 = wdisplays`
`custom_scr_up_cmd_2 = brightnessctl --device=intel_backlight set +1%`
`custom_scr_dn_cmd_2 = brightnessctl --device=intel_backlight set 1%-`
`custom_tooltip_cmd_2 = brightnessctl --device=intel_backlight --machine-readable info | cut -d, -f4`

**Coding notes:**
I initially tried to inherit from Launchers but got stuck by LauncherInfo pure abstract struct; then I had either to modify the related code, or create new (well, it's almost a copy of Launchers…) dedicated struct/class and let mainteners decide which way is better ;-).
I tried to follow coding style and indentation found in other files but I may have missed some things.

It has been quite a while I haven't programmed in C++ and I've learnt a lot by reading the existing code and making this!

Many thanks to Wayfire contributors for this great WM!
Cheers!


pacomod:
Sorry… (branching mistake)

src/panel/widgets/customizables.cpp
```diff
@@ -0,0 +1,315 @@
...
+    if (icon_invert)
+    {
+        invert_pixbuf(ptr_pbuff);
+    }
+    set_image_pixbuf(image, ptr_pbuff, scale);
+}
+
+WfCustomizableButton::WfCustomizableButton()
+{
+    /* I tried to set these as class variables (static),
```
ammen99, line 210: wf-options need to be created after loading the config file, which happens before widgets are initialized, but it won't work before that :)

src/panel/widgets/customizables.cpp
```diff
@@ -0,0 +1,315 @@
...
+    return true;
+}
+
+bool WfCustomizableButton::on_click(GdkEventButton *ev)
+{
+    if (ev->type == GDK_BUTTON_RELEASE)
+    {
+        info->execute(ev->button);
+    }
+    if (ev->button == 1 && ev->type == GDK_BUTTON_PRESS)
```
ammen99, line 155: this doesn't seem to be doing anything, maybe we should remove it?

src/panel/widgets/customizables.cpp
```diff
@@ -0,0 +1,315 @@
...
+    info->execute(ev->direction);
+    return true;
+}
+
+bool WfCustomizableButton::on_enter(GdkEventCrossing* ev)
+{
+    std::string output;
+    if (info->cmd_tooltip.length() > 0)
+    {
+        info->execute(info->cmd_tooltip, &output);
```
ammen99, line 173: This is a cool idea, hadn't thought about it, +1

ammen99 (changes requested):
Sorry for the long delay, looks good overall, and makes for an interesting widget.

godalming123:
what about
```
[customizables]
size = auto
spacing = 10
invert=false

uptime_pretext = "uptime: "
uptime_text = 0 ; gets ovewritten by uptime_text_command
umtime_text_command = uptime
uptime_infront_icon = /some/icon/path.jpg
uptime_refresh_speed = 3s
uptime_btn1 = alacritty --execute echo "some text with uptime information"
uptime_tooltip = uptime
```
something similar to polybar modules where you can specify a command to calculate the text by running a command repeatedly. I also think you could have a customizables section this in my opinion improves readability and they could possibly be implemented into the dock as well

PR #91, opened 2021-06-07
Update default panel widgets

soreau:
This includes window-list and volume widgets by default, in addition to the rest.


ammen99:
Should we update the example config file too?

ammen99:
Thanks!

PR #95, opened 2021-06-13
Fix volume icon muted at startup

rastersoft:
During startup, the volume icon is shown as "muted", no matter whether the volume is really at 0 or at maximum value. The problem seems to be that volume_scale.get_target_value() returns the current value, not the desired one.

This patch fixes this by adding a new method that returns the set value instead of the current one. Tried to return in get_target_value() the final value instead of the current one, but breaks changing the volume.

### DIFF
--- a/src/panel/widgets/volume.cpp
+++ b/src/panel/widgets/volume.cpp
@@ -34,7 +34,7 @@ void WayfireVolumeScale::set_target_value(double value)
 
 double WayfireVolumeScale::get_target_value() const
 {
-    return this->current_volume;
+    return this->current_volume.end;
 }
 
 void WayfireVolumeScale::set_user_changed_callback(


soreau:
I'd recommend something like [this](http://ix.io/3pZP) instead.

rastersoft:
> I'd recommend something like [this](http://ix.io/3pZP) instead.

I tried that, and although it fixes the icon problem, the volume control becomes irresponsible: trying to change the volume results in the volume control returning to the original position (in fact, it's exactly what I meant in the last sentence of the commit text).

ammen99:
Lookin over it, seems like the `get_target_value()` function is not returning what it should. We already have the target value inside the `current_value.end` field, so maybe we should just return that in `get_target_value()` and not change anything else?

rastersoft:
You are right: using current_value.end works. I'll update the patch.

rastersoft:
ammen99: modified this patch to use your changes.

ammen99 (approved):
Thanks!

PR #97, opened 2021-06-17
menu: Refresh menu when apps are installed or removed

### LINKED ISSUE #34: menu does not update when apps are added or removed

soreau:
Fixes #34.


ammen99:
Thanks!

PR #98, opened 2021-06-17
menu: Omit buggy code

soreau:
Gtk3 has a bug which causes flickering tooltips when hovering over
menu items in the menu. Don't set tooltips to avoid this bug.


ammen99 (approved):
thanks :)

PR #100, opened 2021-07-14
panel: Implement default logout user interface

soreau:
When a user clicks the logout button in the menu and no command is set,
open a default logout interface.

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -179,7 +179,32 @@
 	</option>
 	<option name="menu_logout_command" type="string">
 		<_short>Menu Logout Command</_short>
+		<_long>If not specified, a default logout interface will be used.</_long>
+		<default></default>
+	</option>
+	<option name="logout_command" type="string">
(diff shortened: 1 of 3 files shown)


soreau:
screenshot

ammen99:
The idea looks good. However: should we not present the window as a layer-shell overlay on top of everything? I think this is the default in other DEs, but please correct me if I'm wrong.

soreau:
> The idea looks good. However: should we not present the window as a layer-shell overlay on top of everything? I think this is the default in other DEs, but please correct me if I'm wrong.

The way it is currently, the logout UI can be minimized or closed and other views can occlude it, and it calls `gtk-window-present` to unminimize and/or raise the view if it the logout button is clicked again. This is perfectly sane behavior to me. What are the benefits to having it a layer-shell always-on-top view that differ from this behavior?

ammen99 (approved):
Looks cool, thanks!

PR #106, opened 2022-01-06
Allow changing the minimum height and width for the menu.

DrewNaylor:
Recently I tried using Wayfire on my PinePhone, but I found that the menu didn't fit on the screen when using 200% DPI scaling in Wayfire. This PR allows the user to set the minimum height and width of the menu's content from Wayfire Config Manager>Panel>Menu. Not sure on the option names, but I named them "menu_min_content_height" and "menu_min_content_width" in the code, respectively, so they're similar to the "set_min_content_*" calls. They're labeled the same thing in WCM, just in Pascal Case and with spaces instead of underscores to match the other options. For the min and max values, I set them to 100 and 5000 respectively, with 100 for the minimum so the menu doesn't become too small to be usable, and 5000 so things don't crash when opening the menu. The only thing with the max value is that the menu may become too large to be usable depending on the display scaling and resolution if the user sets it too high. Not sure how to handle that. The default was set to 500 as that's what was hardcoded previously.

One thing I did notice is that the logout and search buttons are still a little large in 200% scaling, so they may take up space that the menu may need, but this PR doesn't address that and it may just be a result of using 200% on a 96 DPI display for testing.

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -180,10 +180,12 @@
 	<option name="menu_min_content_width" type="int">
 		<_short>Menu Minimum Content Width</_short>
 		<default>500</default>
+		<min>100</min>
 	</option>
 	<option name="menu_min_content_height" type="int">
 		<_short>Menu Minimum Content Height</_short>
 		<default>500</default>
+		<min>100</min>
 	</option>
 	<option name="menu_logout_command" type="string">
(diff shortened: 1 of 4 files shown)


ammen99 (approved):
LGTM, thanks!

DrewNaylor:
You're welcome!

PR #119, opened 2022-03-18
configurable edge offset

NamorNiradnug:


ammen99 (changes requested):
As a whole, LGTM. But please use a longer deacription for the option in the .XML files, you and I know what edge offset is, but I expect some users to be confused. The description should mention this is about distance from edge to show panel when it is hidden.

ammen99 (approved):

ammen99:
Thanks!

PR #137, opened 2022-09-20
Space between icon and label

NamorNiradnug:
Adds spacing between icons and labels in `battery` and `window-list` widgets


ammen99 (approved):
Makes sense, thanks!

PR #139, opened 2022-10-24
fix #138

### LINKED ISSUE #138: [wf-panel] Autohide (well, "autoshow") is broken on the latest wayfire version
Steps to reproduce:
 - autohide is enabled
 - launch `wf-panel`
   - panel appears and doesn't hide (btw it should)
 - move mouse cursor into the panel's area and then out
   - panel hides
 - move mouse cursor back into the screen edge
   - panel doesn't show but expected to show
   
-----
What I found out:
 - changing `-get_allocated_height()` to `0.001 - get_allocated_height()` [here](https://github.com/WayfireWM/wf-shell/blob/master/src/util/wf-autohide-window.cpp#L239) fixes this bug, but it's weird and makes panel not to hide entirely

NamorNiradnug:
Fixes #138

### DIFF
--- a/src/util/wf-autohide-window.cpp
+++ b/src/util/wf-autohide-window.cpp
@@ -259,7 +259,7 @@ void WayfireAutohidingWindow::schedule_hide(int delay)
 
 bool WayfireAutohidingWindow::m_do_show()
 {
-    y_position.animate(0);
+    y_position.animate(y_position + 1, 0);
     update_margin();
     return false; // disconnect
 }
@@ -286,8 +286,9 @@ bool WayfireAutohidingWindow::update_margin()
     {
         gtk_layer_set_margin(this->gobj(),
             get_anchor_edge(position), y_position);
-
-        this->queue_draw(); // XXX: is this necessary?
+        if (get_window())
+            wl_surface_commit(get_wl_surface());
+        this->queue_draw();
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


src/util/wf-autohide-window.cpp
```diff
@@ -286,8 +286,9 @@ bool WayfireAutohidingWindow::update_margin()
     {
         gtk_layer_set_margin(this->gobj(),
             get_anchor_edge(position), y_position);
-
-        this->queue_draw(); // XXX: is this necessary?
+        if (get_window())
```
ammen99, line 289: Does this work reliably? Previously you have said that this doesn't always work.

src/util/wf-autohide-window.cpp
```diff
@@ -286,8 +286,9 @@ bool WayfireAutohidingWindow::update_margin()
...
+            wl_surface_commit(get_wl_surface());
```
ammen99, line 290: It would also be a good idea to add a comment explaining why we do the commit here instead of queue_draw()

src/util/wf-autohide-window.cpp
NamorNiradnug: It works together with the second chance (which is animation start)

src/util/wf-autohide-window.cpp
NamorNiradnug: It's not instead btw, it doesn't work without `queue_draw`

NamorNiradnug:
I think it works
I use wf-panel with the patch for 2 days

ammen99 (approved):
Thanks!

PR #145, opened 2023-03-07
dock: Fix dock startup failure

soreau:
Apparently, this and priv both have a manager and each are different.


ammen99:
Sounds like one of the two managers should be removed completely.

PR #147, opened 2023-04-04
Notifications fixes

NamorNiradnug:
Several fixes:
 - fix broken string hints
 - fix broken actions (they would appear but not do anything on click)
 - fix setting image from `image_path` hint


ammen99 (approved):
I am not very familiar with the code so I trust you have tested this :) Thanks!

PR #148, opened 2023-04-04
"Command output" widgets

### LINKED ISSUE #124: Command output plugin
Wayfire implements a very small number of plugins and I do understand that there's probably a single developer and not a lot of resources/time/desire/will.

Would be nice to have a "Command Output" plugin which allows to run any shell commands and show their output on the panel.

Check the XFCE project for more info, https://docs.xfce.org/panel-plugins/xfce4-genmon-plugin/start

Commands need to be launched
1) just once on login
2) Periodically once per N seconds (configurable)

Each individual plugin should be clickable, so that I could click it and run some corresponding command of my liking.

This will allow to monitor and implement a ton of various features without coding them.

Thank you.

NamorNiradnug:
Fix #124 

TODO:
 - [x] use `Glib::spawn_async` instead of sync version
 - [x] `icon_size` option

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -274,22 +274,27 @@
 		<short>Commands to run</short>
 		<entry prefix="command_output_" type="string" name="command">
 			<_short>Command</_short>
-			<_long>Command to run. Warning: if command is taking long to run it will freeze the entire panel.</_long>
+			<_long>Command to run</_long>
 		</entry>
-		<entry prefix="command_output_icon_" type="string" name="icon">
-			<_short>Icon</_short>
-			<_long>Icon name to show. Left empty to hide icon.</_long>
-			<default>text-x-script</default>
+		<entry prefix="command_output_tooltip_" type="string" name="tooltip_command">
+			<_short>Tooltip Command</_short>
+			<_long>The tooltip command is executed every time the tooltip is shown.</_long>
+			<default/>
 		</entry>
 		<entry prefix="command_output_period_" type="int" name="period">
 			<_short>Period</_short>
 			<_long>
-Command is launched periodically updating the label above the icon.
+The command is launched periodically, updating the text displayed on the panel.
 Set to 0 to only run the command once on startup.
 Set to -1 to only run it by clicking the button.
 			</_long>
 			<default>0</default>
 			<min>0</min>
+		</entry>
+		<entry prefix="command_output_icon_" type="string" name="icon">
+			<_short>Icon</_short>
+			<_long>Icon name to show. Leave empty to hide icon.</_long>
+			<default>text-x-script</default>
 		</entry>
 		<entry prefix="command_output_icon_size_" type="int" name="icon_size">
 			<_short>Icon Size</_short>
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -278,7 +278,7 @@
 		</entry>
 		<entry prefix="command_output_icon_" type="string" name="icon">
 			<_short>Icon</_short>
-			<_long>Icon name to show</_long>
+			<_long>Icon name to show. Left empty to hide icon.</_long>
 			<default>text-x-script</default>
 		</entry>
 		<entry prefix="command_output_period_" type="int" name="period">
@@ -290,6 +290,31 @@ Set to -1 to only run it by clicking the button.
 			</_long>
 			<default>0</default>
 			<min>0</min>
+		</entry>
+		<entry prefix="command_output_icon_size_" type="int" name="icon_size">
+			<_short>Icon Size</_short>
+			<default>32</default>
+			<min>1</min>
+		</entry>
+		<entry prefix="command_output_icon_position_" type="string" name="icon_position">
+			<_short>Icon Position</_short>
+			<desc>
+				<value>top</value>
+				<_name>Above Text</_name>
+			</desc>
+			<desc>
+				<value>bottom</value>
+				<_name>Below Text</_name>
+			</desc>
+			<desc>
+				<value>left</value>
+				<_name>Left</_name>
+			</desc>
+			<desc>
+				<value>right</value>
+				<_name>Right</_name>
+			</desc>
+			<default>left</default>
 		</entry>
 	</option>
 	</group>
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -289,7 +289,7 @@ Set to 0 to only run the command once on startup.
 Set to -1 to only run it by clicking the button.
 			</_long>
 			<default>0</default>
-			<min>0</min>
+			<min>-1</min>
 		</entry>
 		<entry prefix="command_output_icon_" type="string" name="icon">
 			<_short>Icon</_short>
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -262,5 +262,36 @@
 		<min>1</min>
 	</option>
 	</group>
+	<group>
+	<_short>Command Output</_short>
+	<option name="commands_output_max_chars" type="int">
+		<_short>Max Characters To Display</_short>
+		<_long>Max characters of command output to display.</_long>
+		<min>0</min>
+		<default>10</default>
+	</option>
+	<option name="commands" type="dynamic-list" type-hint="dict">
+		<short>Commands to run</short>
+		<entry prefix="command_output_" type="string" name="command">
+			<_short>Command</_short>
+			<_long>Command to run. Warning: if command is taking long to run it will freeze the entire panel.</_long>
+		</entry>
+		<entry prefix="command_output_icon_" type="string" name="icon">
+			<_short>Icon</_short>
+			<_long>Icon name to show</_long>
+			<default>text-x-script</default>
+		</entry>
+		<entry prefix="command_output_period_" type="int" name="period">
+			<_short>Period</_short>
+			<_long>
+Command is launched periodically updating the label above the icon.
+Set to 0 to only run the command once on startup.
+Set to -1 to only run it by clicking the button.
+			</_long>
+			<default>0</default>
+			<min>0</min>
+		</entry>
+	</option>
+	</group>
 	</plugin>
 </wf-shell>
--- a/src/panel/widgets/command-output.cpp
+++ b/src/panel/widgets/command-output.cpp
@@ -8,6 +8,37 @@
 
 #include <gtk-utils.hpp>
 
+static void label_set_from_command(const std::string & command_line,
+    Gtk::Label& label)
+{
+    Glib::Pid pid;
+    int output_fd;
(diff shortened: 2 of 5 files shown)


mrsteve0924:
this would be a cool widget to get implemented.  im on xfce now and use their plugin like this a lot.  in my wayfire set up i had to move those to a tmux window to monitor command output instead

metadata/panel.xml
```diff
@@ -249,5 +249,36 @@
...
+	<option name="commands" type="dynamic-list" type-hint="dict">
+		<short>Commands to run</short>
+		<entry prefix="command_output_" type="string" name="command">
+			<_short>Command</_short>
+			<_long>Command to run. Warning: if command is taking long to run it will freeze the entire panel.</_long>
+		</entry>
+		<entry prefix="command_output_icon_" type="string" name="icon">
+			<_short>Icon</_short>
+			<_long>Icon name to show</_long>
+			<default>text-x-script</default>
```
ammen99, line 269: Dynamic list entries currently don't support default values in wf-config .. It might be a good feature to add there, it would also simplify using the options in the code.

What is necessary would be:
- Extending the entry structure here to also store the default value: https://github.com/WayfireWM/wf-config/blob/master/include/wayfire/config/compound-option.hpp#L41
- Extend the XML parsing to also parse default values: https://github.com/WayfireWM/wf-config/blob/master/src/xml.cpp#L127
- Adjust the logic for building compound options from the .ini file, it currently explicitly checks whether all entries are initialized: https://github.com/WayfireWM/wf-config/blob/master/src/compound-option.cpp#L20
- Add tests to make sure it works correctly and that we don't break it with future changes

Would you mind doing this? It would be the 'proper' way of fixing things, and then the parsing of the dynamic list in the widget code will be much simplified. On the other hand, if you do not have time or desire to fix wf-config .. I guess we could merge the code as-is, with the dynamic list not specified in the XML file (because it doesn't work in all cases currently).

NamorNiradnug: I didn't know `wf-config` lacks the feature. Yes, I will try to implement it then and update this PR afterwards.

ammen99: Cool, thanks :)

After the update, parsing the options should be very simple, see for example expo: https://github.com/WayfireWM/wayfire/blob/master/plugins/single_plugins/expo.cpp#L93

Basically, wf-config should give you directly a vector of the values, no need to match prefixes manually.

src/panel/widgets/command-output.cpp
```diff
@@ -0,0 +1,113 @@
...
+        if (button != buttons.end())
+        {
+            return *button;
+        }
+
+        buttons.emplace_back();
+        buttons.back().name = name;
+        return buttons.back();
+    };
+    for (const auto &option : WayfireShellApp::get().config.get_section("panel")->get_registered_options())
```
ammen99, line 82: Now that the wf-config branch has been merged, this can be made much simpler.

NamorNiradnug:
TODO:
 - [ ] use `Glib::spawn_async`
 - [ ] `icon_size` option

src/panel/widgets/command-output.cpp
```diff
@@ -0,0 +1,121 @@
...
+        Glib::spawn_async_with_pipes("", Glib::shell_parse_argv(
+            command), Glib::SPAWN_DO_NOT_REAP_CHILD | Glib::SPAWN_SEARCH_PATH_FROM_ENVP,
+            Glib::SlotSpawnChildSetup{},
+            &pid, nullptr, &std_out, nullptr);
+        Glib::signal_child_watch().connect([std_out, this] (
+            Glib::Pid pid,
+            int child_status)
+        {
+            FILE *file = fdopen(std_out, "r");
+            char buf[16];
```
ammen99, line 67: Shouldn't we read up to `max_chars_opt+1` symbols?

ammen99 (commented):
Looks quite good aside from the one comment about max buffer size.

src/panel/widgets/command-output.cpp
NamorNiradnug: I decided to make it `max_chars * 4 + 1` so that it'd work correctly with Unicode symbols.

src/panel/widgets/command-output.cpp
```diff
@@ -0,0 +1,176 @@
...
+        Glib::Pid pid;
+        int output_fd;
+        Glib::spawn_async_with_pipes("", Glib::shell_parse_argv(tooltip_command),
+            Glib::SPAWN_DO_NOT_REAP_CHILD | Glib::SPAWN_SEARCH_PATH_FROM_ENVP,
+            Glib::SlotSpawnChildSetup{}, &pid, nullptr, &output_fd, nullptr);
+        Glib::signal_child_watch().connect([=] (Glib::Pid pid, int exit_status)
+        {
+            FILE *file = fdopen(output_fd, "r");
+            Glib::ustring tooltip_text;
+            std::array<char, 16> buffer;
```
ammen99, line 121: Again, I wonder why limit to 16 .. seems especially that the tooltip might have a longer output.
I'd try reading into a buffer with a flexible size (and you could reuse the code above, if you put that into a function).

NamorNiradnug: It reads the entire output, see the loop below.

NamorNiradnug: Although it actually should be wrapped in a function, I'll do that now.

ammen99 (approved):
Looks great now, thanks :)

birdie-github:
A huge thank you for the patch.

It looks like "Period" is specified in seconds but it's not mentioned. Please do.

birdie-github:
One more thing: XFCE's generic monitor plugin allows quite rich output formatting option including setting font, font style, font color, icon, etc. Hopefully you could look into that as well:

https://docs.xfce.org/panel-plugins/xfce4-genmon-plugin/start

That is a _ton_ of work however, so it's really up to you.

What would be really nice to see as the barebones is:

* Being able to set a tooltip
* Being able to set a command which gets executed when the applet is clicked
* Being able to set colors for text fragments inside the applet

In a perfect world it would be great if Wayfire could just use ... XFCE plugins directly. The API is there. XFCE now uses GTK3 which is Wayland compatible.

That will allow you to incorporate a ton of applets basically for free.

NamorNiradnug:
Thanks for your suggestions! Implementing XFCE plugins sound like a cool idea, I'll look into it.

About other features you mentioned: 
 
> * Being able to set a tooltip

It is able to, use `command_output_tooltip_<name>` option.

> * Being able to set a command which gets executed when the applet is clicked

Current implementation executes the command on click and updates the label on the panel (that's the same command that is launched periodically). Are you suggesting to make it a separate command? 

> * Being able to set colors for text fragments inside the applet

It is able to: it uses [Pango Markup](https://docs.gtk.org/Pango/pango_markup.html).

birdie-github:
> Current implementation executes the command on click and updates the label on the panel (that's the same command that is launched periodically). Are you suggesting to make it a separate command?

Yeah, that's my thinking but I just want to avoid confusion. It won't be nice if clicking the applet could either update it or run an external command, e.g. a terminal application with `top` in it. So, yeah, we are talking about two different commands, the primary one that generates the applet output and the second one which gets executed when it's called specifically.

XFCE always runs the second command on click, and refreshing requires right mouse clicking and selecting "Update Now" from a pop-up menu. Without a pop-up menu I've now idea how to set them apart. Maybe using middle mouse click?

Anyways, it's all completely up to you. I'm OK if you disregard my request altogether :-)

mrsteve0924:
hello, trying to implement this.    do i add 'command-output' to 'widgets_x =' in wf-shell.ini? 

and then where to add the command to run?  thanks

NamorNiradnug:
> hello, trying to implement this. do i add 'command-output' to 'widgets_x =' in wf-shell.ini?
> 
> and then where to add the command to run? thanks

Hi. Yes, you add `command-output` to `widgets_x`. Then for each command you can specify the following options:
```
command_output_<name> = <command>  # this one is required
command_output_icon_<name> = <icon-name>
command_output_icon_size_<name> = <size>
# ...
```
See full list of available options in the `metadata/panel.xml` file.

mrsteve0924:
thanks for the reply.  looks good.  nice work!

any way to increase the max text size. i set it for 60 but looks like it stops at around 30 then adds ...

mrsteve0924:
thanks for the reply.  looks good.  nice work!

soreau:
> thanks for the reply. looks good. nice work!
> 
> 

This does look nice.. how can I get the weather applet like in this screenshot?

mrsteve0924:
curl -s http://wttr.in/?format=2

check out https://github.com/chubin/wttr.in

soreau:
> curl -s http://wttr.in/?format=2
> 
> check out https://github.com/chubin/wttr.in

Thanks!

rodcps:
> curl -s http://wttr.in/?format=2
> 
> check out https://github.com/chubin/wttr.in


I need help using the command-output plugin. I'm a humanities major and I've tried every way to make it work. I'd like to bring up the weather forecast, but I haven't been successful!

I'm editing the wf-shel.ini file and adding the code below:

widgets_center = command-output=curl -H "Accept-Language: pt-br" 'wttr.in/Campinas?format=%l:+%C+%t'

The command output in the bar should be:
Campinas: Overcast +XX°C⏎

But I can't get it to work. I've tried everything. If you can help me with this, I'd appreciate it!

mrsteve0924:
you need add 'command-output' to the widget line by itself
widgets_center = command-output

then add your curl statement  and other parameters you need to the 'command_output_ ' variable .  Here is an example:
command_output_wttr = curl -H "Accept-Language: pt-br" 'wttr.in/Campinas?format=%l:+%C+%t'
command_output_period_wttr = 1200
command_output_tooltip_wttr = echo weather
command_output_icon_wttr =
command_output_icon_size_wttr = 25

rodcps:
> you need to add 'command-output' to the widget line by itself widgets_center = command-output
> 
> then add your curl statement and other parameters you need to the 'command_output_ ' variable . Here is an example: command_output_wttr = curl -H "Accept-Language: pt-br" 'wttr.in/Campinas?format=%l:+%C+%t' command_output_period_wttr = 1200 command_output_tooltip_wttr = echo weather command_output_icon_wttr = command_output_icon_size_wttr = 25

------
Thank you very much mrsteve0924! It works now and it looks really cool, just one point to contribute is the line that defines the character size (commands_output_max_chars = 25) which in my case was defined as 2 only, in the end my code looked like this:

commands_output_max_chars = 25
widgets_right = command-output tray volume clock
command_output_wttr = curl wttr.in/Campinas?format=%c%t+%h+%p
command_output_period_wttr = 1200
command_output_tooltip_wttr = echo Time
command_output_icon_wttr =
command_output_icon_size_wttr = 24

Here is the screenshot and thank you very much again!

20241201_08h30m59s_grim

PR #153, opened 2023-07-07
Implement StatusNotifierItem

### LINKED ISSUE #29: add status notifier item in wf-panel
as systemtray is not supported on wayland.
but sni is supposed to be supported.
add the sni widget in wf-panel.

https://freedesktop.org/wiki/Specifications/StatusNotifierItem/

NamorNiradnug:
Fix #29

WIP features include:
- [x] handling SNI's `New*` signals
- [x] working properly with `blueman-applet` (~~currently I have no idea what's wrong~~ it was a missing `invocation->return_value(...)` call)
- [x] config options


ammen99 (approved):
My knowledge of SNI is very limited, so I have not reviewed the PR in detail. But it looks good, I also tested it and it seems to work, at least with VLC. Great work and thanks for the PR!

ammen99:
Hmm, @NamorNiradnug, nm-applet does not show up here, any ideas why that might be the case?

PR #154, opened 2023-07-08
Fix #109

### LINKED ISSUE #109: [wf-dock] Apply auto hide on startup (feature request)
If I set auto-hide it works as expected, but only after my mouse has hovered over it once. On login it appears on all displays (as expected) but remains visible until I've moved my mouse over it on each display. 

I'd like it to either start hidden  or start visible and auto-hide after the specified interval regardless of whether I hover my mouse over it.

NamorNiradnug:
fixes #109

### DIFF
--- a/src/util/wf-autohide-window.cpp
+++ b/src/util/wf-autohide-window.cpp
@@ -53,11 +53,11 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
 
     if (output->output)
     {
-        const auto* listener = new zwf_output_v2_listener {
+        static const zwf_output_v2_listener listener = {
             .enter_fullscreen = [](void* data, zwf_output_v2 *){ ((WayfireAutohidingWindow*)data)->increase_autohide(); },
             .leave_fullscreen = [](void* data, zwf_output_v2 *){ ((WayfireAutohidingWindow*)data)->decrease_autohide(); }
         };
-        zwf_output_v2_add_listener(output->output, listener, this);
+        zwf_output_v2_add_listener(output->output, &listener, this);
     }
 }
 
--- a/src/util/wf-autohide-window.cpp
+++ b/src/util/wf-autohide-window.cpp
@@ -50,6 +50,15 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
             if (this->active_button)
                 unset_active_popover(*this->active_button);
         });
+
+    if (output->output)
+    {
+        const auto* listener = new zwf_output_v2_listener {
+            .enter_fullscreen = [](void* data, zwf_output_v2 *){ ((WayfireAutohidingWindow*)data)->increase_autohide(); },
+            .leave_fullscreen = [](void* data, zwf_output_v2 *){ ((WayfireAutohidingWindow*)data)->decrease_autohide(); }
+        };
+        zwf_output_v2_add_listener(output->output, listener, this);
+    }
 }
 
 WayfireAutohidingWindow::~WayfireAutohidingWindow()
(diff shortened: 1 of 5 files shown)


ammen99 (commented):
I am confused as to how this works and why it is necessary, would you mind elaborating?

NamorNiradnug:
> I am confused as to how this works and why it is necessary, would you mind elaborating?

The reason the panel won't hide on startup is that [this](https://github.com/WayfireWM/wf-shell/blob/master/src/util/wf-autohide-window.cpp#L91) check fails, because `autohide_counter` is initialized with zero. So, as I understand what `autohide_counter` is, it should be incremented.

NamorNiradnug:
Oh, I see the problem. It completely ignores if autohide is enabled or not. Maybe then `autohide_counter` should be incremented only if autohide is enabled?

ammen99:
Your new approach certainly makes more sense :) I'm wondering whether it would make sense to instead make the generic autohide window handle the autohide options as well? I don't think the panel or the dock care about the specific value of the option, so we could share the code, do you think this is feasible?

NamorNiradnug:
There are two problems currently
1. `wf-dock` is shown on the left instead of being centred when its `autohide` option is `false` and `exclusive_zone` is set
2. `wf-panel` hides automatically when a view on the corresponding output goes fullscreen. Shouldn't the feature also be shared between panel and dock as well?

ammen99:
> 1. `wf-dock` is shown on the left instead of being centred when its `autohide` option is `false` and `exclusive_zone` is set

Hmm, sounds like this could be a bug on Wayfire's side. I'll try to look into it tomorrow.



> 2\. `wf-panel` [hides automatically](https://github.com/WayfireWM/wf-shell/blob/master/src/panel/panel.cpp#L307) when a view on the corresponding output goes fullscreen. Shouldn't the feature also be shared between panel and dock as well?

Yes, sounds good. IIRC the dock used to always auto-hide, which is why the feature wasn't necessary before.

src/util/wf-autohide-window.cpp
```diff
@@ -50,6 +50,15 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
             if (this->active_button)
                 unset_active_popover(*this->active_button);
         });
+
+    if (output->output)
+    {
+        const auto* listener = new zwf_output_v2_listener {
+            .enter_fullscreen = [](void* data, zwf_output_v2 *){ ((WayfireAutohidingWindow*)data)->increase_autohide(); },
```
ammen99, line 57: Woah, didn't know you can do that (passing lambda to wayland callbacks), cool :)

ammen99:
> Hmm, sounds like this could be a bug on Wayfire's side. I'll try to look into it tomorrow.

I found the problem, Wayfire's anchored areas do not center the surface properly with exclusive zone. I have a few other changes planned in that respect, so I'll create a Wayfire PR soon.

ammen99:
@NamorNiradnug https://github.com/WayfireWM/wayfire/pull/1829 should fix the issue with the dock placement, can you confirm?

ammen99 (approved):
LGTM, should we merge?

NamorNiradnug:
There is a memory leak, because `listener` is never deleted, I'll fix it soon, please don't merge it yet.

NamorNiradnug:
> @NamorNiradnug [WayfireWM/wayfire#1829](https://github.com/WayfireWM/wayfire/pull/1829) should fix the issue with the dock placement, can you confirm?

Yep, it works!

NamorNiradnug:
> LGTM, should we merge?

I think it's ready to be merged now.

ammen99:
Thanks!

PR #156, opened 2023-07-09
refactor window-list

### LINKED ISSUE #150: High CPU usage in windowlist when it's too wide
**To Reproduce**
1. Use wf-panel
2. Use one panel on bottom of screen with several widgets including a window list,
3. Create lots of windows, such that the window list width is constrained by available panel width

**Expected behavior**
Windows icons in list get smaller and CPU usage stays low.

**Actual behaviour**
wf-panel starts consuming 100% of one CPU. WayfireWindowList::on_draw continuously runs this code - maybe once every frame?

```
    /* We have changed the size/number of toplevels. On top of that, our list
     * is longer that the max size, so we need to re-layout the buttons */
    if (preferred_width > allocated_width && toplevels.size() > 0)
        set_button_width(get_target_button_width());
```

**Wayfire version**
0.7.4-2 from debian testing

**Panel configuration**
```
widgets_left = menu launchers spacing4 window-list
widgets_center = none
widgets_right = volume network battery clock
# ... configuration for 6 launchers
launchers_spacing = 0
launchers_size = 34
```

NamorNiradnug:
Probably fixes #150

This patch refactors the calculation of `WayfireToplevel`'s width. Instead of recalculating the width manually, it simply does `label.set_ellipsize` and `label.set_max_width_chars`.

Also there are a few minor style-ish changes suggested by clang-tidy.

### DIFF
--- a/src/panel/widgets/window-list/toplevel.cpp
+++ b/src/panel/widgets/window-list/toplevel.cpp
@@ -53,12 +53,16 @@ class WayfireToplevel::impl
         zwlr_foreign_toplevel_handle_v1_add_listener(handle,
             &toplevel_handle_v1_impl, this);
 
+        static constexpr auto SIZE_LIMIT_COEF = 0.01;
+        Gdk::Rectangle monitor_rect;
+        window_list->output->monitor->get_geometry(monitor_rect);
+        label.set_max_width_chars(monitor_rect.get_width() * SIZE_LIMIT_COEF);
+        label.set_ellipsize(Pango::EllipsizeMode::ELLIPSIZE_END);
         button_contents.add(image);
         button_contents.add(label);
         button_contents.set_halign(Gtk::ALIGN_START);
         button_contents.set_spacing(5);
         button.add(button_contents);
-        button.set_tooltip_text("none");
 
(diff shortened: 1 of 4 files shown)


NamorNiradnug:
Known problem: the `WayfireToplevel` widget becomes too small when all the toplevels have a short titles.

lcolitti (commented):
This commit simplifies things but it doesn't fix the problem. Once there are too many windows, the box starts growing, and moving to the right any widgets that are to the right of it.

For example, to the right of the window list, I have volume, network, battery, tray, and clock widgets, and if I open too many windows, the window list starts growing and pushing those widgets off the screen.

NamorNiradnug:
> This commit simplifies things but it doesn't fix the problem. Once there are too many windows, the box starts growing, and moving to the right any widgets that are to the right of it.
> 
> For example, to the right of the window list, I have volume, network, battery, tray, and clock widgets, and if I open too many windows, the window list starts growing and pushing those widgets off the screen.

Hi
Yeah, I know there's the problem, but that's why the PR is still a draft)
Thanks for your review anyway!

PR #157, opened 2023-07-14
tray: handle property-not-found error in `fetch_property`

NamorNiradnug:
Fixes the bug that caused `nm-applet`'s icon to not be updated.


NamorNiradnug:
Closing in favor of #159

Issue #35, opened 2019-11-10
add categories in menu widget


ghost:
add categories in menu.
and
it should be able enable and disable categorisation of menu widget with wf-shell.ini

ammen99:
fixed in #230

PR #160, opened 2023-07-22
refactor menu widget

### LINKED ISSUE #111: [wf-panel] Start-menu: Allow to select/launch items via keyboard
Often times I type in a search term, and then instinctively use my `down` key to select the item and launch it with `enter`.

But it seems that is not implemented.

My proposed idea is, that the first item should always be auto selected by default, and the user can use his arrow keys to change the selected item.

The second option is, that the user first needs to press the down arrow to enter the item list. (But frankly I don't think that extra step is of any use for the user.)

### LINKED ISSUE #112: [wf-panel] Discard search field text after closing start menu
Whenever you type a search term into the start menu, this text persists unless you delete it manually.

I think that this text should be discarded when the menu is closed, and the user should be presented with a fresh menu.

NamorNiradnug:
TODO:
 - [ ] Fix #111 
 - [x] Fix #112

Maybe:
 - [ ] Implement #35

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -14,41 +14,24 @@
 #include "launchers.hpp"
 #include "wf-autohide-window.hpp"
 
-#define MAX_LAUNCHER_NAME_LENGTH 11
 const std::string default_icon = ICONDIR "/wayfire.png";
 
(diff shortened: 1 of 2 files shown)


ammen99:
I believe most of the items here have been addressed in #230 but nonetheless thanks for the pr :)

PR #162, opened 2023-07-29
icon size option for volume and notifications

NamorNiradnug:
- Adds icon size option for `volume` and `notifications` widgets
 - Changes `battery_icon_size`'s default value to `32`
 - Adds missing `*_icon_size`'s min value


ammen99 (approved):
Seems simple enough so lgtm, thanks!

PR #168, opened 2023-08-03
Notifications: make daemon a singleton

NamorNiradnug:
Fixes an awful mistake: `Daemon` was a namespace instead of singleton. 
Fixes that notifications center disappears after widgets reloading.


ammen99 (approved):
LGTM, feel free to merge if you don't have anything else to add ;)
As usual, thanks for your contributions to wf-shell, it is great to see so many improvements in a short time!

PR #171, opened 2023-08-03
volume: don't show-and-hide popover on init

NamorNiradnug:
It's a very annoying thing because changing the list of widgets in `wmc` makes panel capture the focus.


ammen99 (approved):
LGTM!

PR #173, opened 2023-08-03
clock: use pack_start to be placed correctly between other widgets

### LINKED ISSUE #135: Some of the panel gadgets can't place to the correct position...
when I am trying to have left : clock volume

it doesn't working...

some not often choosed positioning orders aren't working too!

hope you can fix this..!

NamorNiradnug:
Fixes #135 

Currently `clock` widget is always placed at right regardless of its position in config file (e.g. `battery clock` or `clock battery` produces the same result). IMHO, this behavior seems weird.


ammen99 (approved):
Makes sense, we should use the same function everywhere.

PR #175, opened 2023-08-04
panel: stringify network_ and battery_status

NamorNiradnug:
This makes config files more readable.

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -104,21 +104,19 @@
 	</group>
 	<group>
 	<_short>Battery</_short>
-	<option name="battery_status" type="int">
+	<option name="battery_status" type="string">
 		<_short>Battery Status</_short>
-		<default>1</default>
-		<min>0</min>
-		<max>2</max>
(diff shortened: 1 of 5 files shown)


NamorNiradnug:
By the way, I've noticed that displaying IP is not implemented, so maybe the corresponding option should be removed, at least until it's implemented?

Tbh, I'm not sure if this is a really useful feature, but I think I could implement this if you think otherwise.

ammen99:
@NamorNiradnug Feel free to remove any unused code or unsupported values :)

soreau:
Not related to this PR, but wf-background doesn't use [this variable](https://github.com/WayfireWM/wf-shell/blob/master/src/background/background.hpp#L44) though it's set [here](https://github.com/WayfireWM/wf-shell/blob/master/src/background/background.cpp#L286).

ammen99 (approved):
I don't like breaking changes but the current interface is indeed very confusing, so LGTM.

Thanks!

PR #178, opened 2023-08-05
volume: hide popover after scrolling

### LINKED ISSUE #177: volume icon slider doesn't disappear
scrolling over volume widget in panel will activate a slider but it doesn't disappear on its own.  volume icon has to be clicked for it to disappear. if you don't click it away it holds on to focus. 

slider disappears if the volume is changed from the keyboard or from terminal commands

NamorNiradnug:
Fixes #177 

Tbh, I have no idea how and why it works.


NamorNiradnug:
> Sounds as if the actual problem is that the timeout in line 107 is not actually set, because the popup grabs input .. Do we actually ever want the popup to stay forever on screen and grab the keyboard? I'd say the popup should always auto-hide after a configurable timeout and never grab the keyboard.
> 
> Currently, it grabs the keyboard so that clicking on another view will auto-close the popup (which is actually broken in recent times, I am starting to work on that) - but that won't be a problem at all if we always auto-close after, say, 2 seconds of inactivity.

Sounds like a good feature, but I think it should autohide faster after changing volume level via scrolling or pressing keyboard button because 2 seconds (or whatever, it obviously should be configurable) it too much to simply see the current volume level.

ammen99:
> > Sounds as if the actual problem is that the timeout in line 107 is not actually set, because the popup grabs input .. Do we actually ever want the popup to stay forever on screen and grab the keyboard? I'd say the popup should always auto-hide after a configurable timeout and never grab the keyboard.
> > Currently, it grabs the keyboard so that clicking on another view will auto-close the popup (which is actually broken in recent times, I am starting to work on that) - but that won't be a problem at all if we always auto-close after, say, 2 seconds of inactivity.
> 
> Sounds like a good feature, but I think it should autohide faster after changing volume level via scrolling or pressing keyboard button because 2 seconds (or whatever, it obviously should be configurable) it too much to simply see the current volume level.

A configuration option sounds good ;)

ammen99:
Seems like the original issue has been fixed somewhere on the way to gtk4, or do you still experience it?

PR #183, opened 2023-09-30
use `tray` and `notifications` widgets by default

### LINKED ISSUE #182: Add new widgets by default
I think we should add the tray widget at the minimum, to wf-shell.ini.example.

NamorNiradnug:
Closes #182

### DIFF
--- a/wf-shell.ini.example
+++ b/wf-shell.ini.example
@@ -13,13 +13,13 @@ randomize = 0
 [panel]
 # widgets_* is a space-separated list of widgets to be displayed
 # at the corresponding part of the panel
-# Supported widgets are: launchers clock network battery window-list volume menu notifications
+# Supported widgets are: launchers clock network battery window-list volume menu notifications tray command-output
 # A special widgets is spacing widgets, it can be used to add padding everywhere on the panel
 # To use it, just append the amount of pixels you want as a padding
 # to the word "spacing" and use it as a plugin
 widgets_left = menu spacing4 launchers window-list
 widgets_center = none
-widgets_right = volume network battery clock
+widgets_right = tray notifications volume network battery clock
 
 # The minimal size of the panel. Note that some widgets might force panel bigger than this size.
 # All widgets also have individual settings for size
--- a/wf-shell.ini.example
+++ b/wf-shell.ini.example
@@ -19,7 +19,7 @@ randomize = 0
 # to the word "spacing" and use it as a plugin
 widgets_left = menu spacing4 launchers window-list
 widgets_center = none
-widgets_right = tray notifications volume network battery clock
+widgets_right = command-output tray notifications volume network battery clock
 
 # The minimal size of the panel. Note that some widgets might force panel bigger than this size.
 # All widgets also have individual settings for size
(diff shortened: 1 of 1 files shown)


soreau:
Should we maybe enable command widget with sensors output or something?

NamorNiradnug:
> Should we maybe enable command widget with sensors output or something?

That's a great idea, I think the easiest way is to `grep` it from `sensors`, but I'm not really good at shell stuff, could you please help with the command?

soreau:
> > Should we maybe enable command widget with sensors output or something?
> 
> That's a great idea, I think the easiest way is to `grep` it from `sensors`, but I'm not really good at shell stuff, could you please help with the command?

Yes, can you say what criteria the command output should have or how to use commands widget? There aren't any docs but the source so far.

soreau:
So far, I've added `command-output` to `widgets_right` in ~/.config/wf-panel.ini and set `[panel]` `commands = sensors`, but it doesn't even show in the panel. I also tried restarting the panel.

ammen99:
> That's a great idea, I think the easiest way is to `grep` it from `sensors`, but I'm not really good at shell stuff, could you please help with the command?

```
sensors | grep Package | cut -d ' ' -f 5
```

This command outputs the current temperature of the CPU, at least on my system .. I hope it works similarly for you :)

What it does is look for the line containing `Package id 0:` (it may be called a bit differently on your system, idk) and then selects the 5th field on that line when separating by spaces - and this fifth field is the temperature.

NamorNiradnug:
> So far, I've added `command-output` to `widgets_right` in ~/.config/wf-panel.ini and set `[panel]` `commands = sensors`, but it doesn't even show in the panel. I also tried restarting the panel.

It works with `command_output_sensors = "sensors"`
And here is how it looks like :)
image

NamorNiradnug:
> > That's a great idea, I think the easiest way is to `grep` it from `sensors`, but I'm not really good at shell stuff, could you please help with the command?
> 
> ```
> sensors | grep Package | cut -d ' ' -f 5
> ```
> 
> This command outputs the current temperature of the CPU, at least on my system .. I hope it works similarly for you :)
> 
> What it does is look for the line containing `Package id 0:` (it may be called a bit differently on your system, idk) and then selects the 5th field on that line when separating by spaces - and this fifth field is the temperature.

Unfortunately it doesn't work on my machine... I'm afraid the names of sensors are different on each machine, and we need something that would work out of box for most users.

ammen99:
> > > That's a great idea, I think the easiest way is to `grep` it from `sensors`, but I'm not really good at shell stuff, could you please help with the command?
> > 
> > 
> > ```
> > sensors | grep Package | cut -d ' ' -f 5
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This command outputs the current temperature of the CPU, at least on my system .. I hope it works similarly for you :)
> > What it does is look for the line containing `Package id 0:` (it may be called a bit differently on your system, idk) and then selects the 5th field on that line when separating by spaces - and this fifth field is the temperature.
> 
> Unfortunately it doesn't work on my machine... I'm afraid the names of sensors are different on each machine, and we need something that would work out of box for most users.

I think we can at most add a commented out example for one of the sensors, just so that interested users can see that it is possible .. And I have the feeling most users who are willing to read the ini file will also have the know-how to figure out which command they need exactly :)

soreau:
Also, I would like to be able to click or hover on the button to see the full output.

ammen99:
> Also, I would like to be able to click or hover on the button to see the full output.

That is what `command_output_tooltip_<x>` is for.

soreau:
> > Also, I would like to be able to click or hover on the button to see the full output.
> 
> That is what `command_output_tooltip_<x>` is for.

Ah, thanks.

ammen99:
> I built this command which works in terminal: `sensors|grep Core|head -n 1|awk '{print $3 $4 $5 $6 $7 $8 $9}'` but when I set `command_output_sensors = sensors|grep Core|head -n 1|awk '{print $3 $4 $5 $6 $7 $8 $9}'`, nothing happens. When I restart the panel, it crashes with `what : Failed to execute child process “sensors|grep” (No such file or directory)`.

I think it might be a good idea to execute not just the command given in the config file, but `/bin/sh -c "<the command>"`, like [Wayfire does](https://github.com/WayfireWM/wayfire/blob/master/src/core/core.cpp#L399) - this will definitely make writing commands easier.

Also, it would indeed be nice if the panel doesn't crash when the command fails ..

soreau:
The tooltip is not updated with the same frequency as the button, when hovering over constantly.

soreau:
But all in all, this commands widget is a great idea.

soreau:
Also, the tray widget seems to work great, no problems there.

soreau:
Commands with color codes like weather.. it shows the color code instead of applying the code.

ammen99:
I added a few more options to the config file.

soreau:
With regards to command-output, I set it all up then changed a setting in wcm->panel and it erased all the command-output settings.

PR #186, opened 2023-10-07
Don't autohide if zwf_shell_manager_v2 is not supported by compositor

### LINKED ISSUE #184: dock can't show up after it hides
### reproduce steps
1. Enable the `autohide` option in dock section of `wf-shell.ini`
2. Start `wf-dock` and wait for the dock to hide
3. Try to hover on the dock edge, it can't show up

soreau:
Without wayfire-shell plugin loaded, wf-panel and wf-dock autohide and never show again. If zwf_shell_manager_v2 is not adavertised, don't autohide and throw an error.

Fixes #184.

### DIFF
--- a/src/util/wf-autohide-window.cpp
+++ b/src/util/wf-autohide-window.cpp
@@ -30,20 +30,8 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
     this->position.set_callback([=] () { this->update_position(); });
     this->update_position();
 
-    this->edge_offset.set_callback([=] () { this->setup_hotspot(); });
-
-    this->autohide_opt.set_callback([=] { update_autohide(); });
-    set_auto_exclusive_zone(!autohide_opt);
-
     this->signal_draw().connect_notify(
         [=] (const Cairo::RefPtr<Cairo::Context>&) { update_margin(); });
-
-    this->signal_size_allocate().connect_notify(
-        [=] (Gtk::Allocation&)
-    {
-        this->set_auto_exclusive_zone(this->has_auto_exclusive_zone);
-        this->setup_hotspot();
-    });
 
     this->signal_focus_out_event().connect_notify(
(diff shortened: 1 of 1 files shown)


src/util/wf-autohide-window.cpp
```diff
@@ -54,20 +42,39 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
...
-            }
-        };
-        zwf_output_v2_add_listener(output->output, &listener, this);
+
+    if (!output->output)
+    {
+        std::cerr << "WARNING: Compositor does not support zwf_shell_manager_v2 " << \
+            "disabling hotspot and autohide features " << \
+            "(is wayfire-shell plugin enabled?)" << std::endl;
+        increase_autohide();
```
ammen99, line 51: increase_autohide? Sounds like it should decrease_autohide(), if anything should be done about it at all ..

src/util/wf-autohide-window.cpp
```diff
@@ -54,20 +42,39 @@ WayfireAutohidingWindow::WayfireAutohidingWindow(WayfireOutput *output,
...
+        [=] (Gtk::Allocation&)
+    {
+        this->set_auto_exclusive_zone(this->has_auto_exclusive_zone);
+        this->setup_hotspot();
+    });
+
+    this->edge_offset.set_callback([=] () { this->setup_hotspot(); });
+
+    this->autohide_opt.set_callback([=] { update_autohide(); });
+    set_auto_exclusive_zone(!autohide_opt);
```
ammen99, line 65: I believe exclusive zone is important if we do not have autohide enabled ..

ammen99 (commented):
I think it would be a little bit simpler to change the code so that it does not read the autohide_opt directly, but instead a wrapper function `get_autohide_enabled()` which just returns false regardless of the option, if the output is not set.

ammen99 (approved):
Thanks :)

PR #188, opened 2023-10-07
Unbreak build on FreeBSD

jbeich:


ammen99 (approved):
Thanks :) I'll add those to the 0.8.0 release as well ..

Issue #198, opened 2023-11-24
Regression: composer allocating windows in the wf-panel's area


lavender-agatha:
Since the latest release of wf-shell, I am experiencing an issue with my composer (wayfire).

Whenever I maximize a window or dock it using the grid plugin, the composer seems to be including the wf-panel's area into the area where it puts the window, hiding part of the window.

I checked this behavior and it's not only happening in the latest release, but also in the git version. 

So I bisected it and found out this regression was introduced in 665263593350fddd92322e553fc09895cdebf209, as it changes the auto-hiding code and I found out that it works just fine in the previous commit 0fb7f50d41856a5ee3668286b7362cdb50528b5e. In the end I'm adding two small videos showing the behavior in both cases.

I tested it with both my current config file, the example config file, etc. I have auto-hiding disabled.

I am reading the code to try to understand it and work on a fix.

Here are the videos demonstrating the behaviors:

Behavior in 665263593350fddd92322e553fc09895cdebf209:
https://github.com/WayfireWM/wf-shell/assets/24969130/5761faf1-a142-4b09-b30d-303ffdbcf8ec

Behavior in 0fb7f50d41856a5ee3668286b7362cdb50528b5e:
https://github.com/WayfireWM/wf-shell/assets/24969130/c386601f-6a2a-4cab-bf27-e302377326c1

lavender-agatha:
Important detail: this bug only happens when the shell plugin isn't loaded.

Yet, it's still an unnecessary dependency that could affect wf-shell users running a different compositor.

I'm making a PR to fix it.

ammen99:
fixed in #199

PR #199, opened 2023-11-29
Fix excusive zone when `zwf_shell_manager_v2` is unavailable

lavender-agatha:
Commit 6652635 fixed autohide issues when `zwf_shell_manager_v2` isn't available, but changed, as a side effect, the behaviour of the window's `auto_exclusive_zone`, causing an unnecessary dependency on the `zwf_shell_manager_v2` for the exclusive zone as well, affecting wf-panel when the compositor's wayfire-shell plugin is disabled (see issue #198 for details).

Fix it by refactoring the `WayfireAutohidingWindow` class, dividing some methods charged with too much roles (i.e the
`WayfireAutohidingWindow::set_autoexclusive_zone()`), transfering constructor code into new methods, introducing some new logic controls both in the class body and in the callbacks.

### Testing
To ensure everything is working fine, I did several tests on both `wf-panel` and `wf-deck` to check the behavior under different conditions. Here's what I tested:

#### `wf-panel`
I tested it both with the wayfire-shell plugin enabled and disabled, and changing the autohide config on the fly, both by starting it enabled and disabled. I looked for some key behaviors, and it worked as expected (by me, at least) on different scenarios. Here I'm pasting my notes on it while I tested:

S->wayfire-shell plugin loaded at beginning
I->initial autohide value
C{1,2}->Changes of autohide value 1 and 2

Test	Status	S	I	C1	C2
1	Success	0	0	1	0
2	Success	0	1	0	1
3	Success	1	0	1	0
4	Success	1	1	0	1

Results:

Test Scenario Id #1
[y] Show zwf_shell_manager_v2 disabled warning at startup
[y] Reposition windows out of panel's area at startup
[n] Actually enable autohide and disable exclusive zone when autohide is set to true
[y] Show swf_shell_manager_v2 disabled warning when attempts to enable autohide
[-] Reposition windows on panel's area when enable autohide
[-] Reposition windows out of pane's area when disable autohide

Test Scenario Id #2
[y] Show zwf_shell_manager_v2 disabled warning at startup
[y] Reposition windows out of panel's area at startup
[n] Actually enable autohide and disable exclusive zone when autohide is set to true
[y] Show swf_shell_manager_v2 disabled warning when attempts to enable autohide
[-] Reposition windows on panel's area when enable autohide
[-] Reposition windows out of pane's area when disable autohide

Test Scenario Id #3
[n] Show zwf_shell_manager_v2 disabled warning at startup
[y] Reposition windows out of panel's area at startup
[y] Actually enable autohide and disable exclusive zone when autohide is set to true
[n] Show swf_shell_manager_v2 disabled warning when attempts to enable autohide
[y] Reposition windows on panel's area when enable autohide
[y] Reposition windows out of pane's area when disable autohide (*)

Test Scenario Id #4
[n] Show zwf_shell_manager_v2 disabled warning at startup
[n] Reposition windows out of panel's area at startup
[y] Actually enable autohide and disable exclusive zone when autohide is set to true
[n] Show swf_shell_manager_v2 disabled warning when attempts to enable autohide
[y] Reposition windows on panel's area when enable autohide
[y] Reposition windows out of pane's area when disable autohide (*)

(*) Sometimes it needs to move the mouse over the edge to make the panel visible again

#### `wf-dock`
Since here there aren't as much factors as in `wf-panel`, I just tested it by opening it with and without the `zwf_shell_manager_v2` connected. In both cases it behaved as expected, with autohiding enabled when the shell manager is available, and disabled when not.

### DIFF
--- a/src/util/wf-autohide-window.hpp
+++ b/src/util/wf-autohide-window.hpp
@@ -89,11 +89,15 @@ class WayfireAutohidingWindow : public Gtk::Window
     int last_edge_offset = -1;
 
     WfOption<bool> autohide_opt;
+    bool autohide_enabled = autohide_opt;
     bool last_autohide_value = autohide_opt;
+    void setup_autohide();
     void update_autohide();
 
-    bool has_auto_exclusive_zone = false;
-    int last_zone = 0;
+    bool auto_exclusive_zone = !autohide_enabled;
+    int auto_exclusive_zone_size = 0;
+    void setup_auto_exclusive_zone();
+    void update_auto_exclusive_zone();
 
     sigc::connection pending_show, pending_hide;
     bool m_do_show();
@@ -107,6 +111,7 @@ class WayfireAutohidingWindow : public Gtk::Window
     bool input_inside_panel     = false;
     zwf_hotspot_v2 *edge_hotspot  = NULL;
     zwf_hotspot_v2 *panel_hotspot = NULL;
+    bool zwf_support = false;
     std::unique_ptr<WayfireAutohidingWindowHotspotCallbacks> edge_callbacks;
     std::unique_ptr<WayfireAutohidingWindowHotspotCallbacks> panel_callbacks;
     void setup_hotspot();
--- a/src/util/wf-autohide-window.hpp
+++ b/src/util/wf-autohide-window.hpp
@@ -111,7 +111,6 @@ class WayfireAutohidingWindow : public Gtk::Window
     bool input_inside_panel     = false;
     zwf_hotspot_v2 *edge_hotspot  = NULL;
     zwf_hotspot_v2 *panel_hotspot = NULL;
-    bool zwf_support = false;
     std::unique_ptr<WayfireAutohidingWindowHotspotCallbacks> edge_callbacks;
     std::unique_ptr<WayfireAutohidingWindowHotspotCallbacks> panel_callbacks;
     void setup_hotspot();
--- a/src/util/wf-autohide-window.hpp
+++ b/src/util/wf-autohide-window.hpp
@@ -89,12 +89,11 @@ class WayfireAutohidingWindow : public Gtk::Window
     int last_edge_offset = -1;
 
     WfOption<bool> autohide_opt;
-    bool autohide_enabled = autohide_opt;
     bool last_autohide_value = autohide_opt;
     void setup_autohide();
     void update_autohide();
 
-    bool auto_exclusive_zone = !autohide_enabled;
+    bool auto_exclusive_zone = !autohide_opt;
     int auto_exclusive_zone_size = 0;
     void setup_auto_exclusive_zone();
     void update_auto_exclusive_zone();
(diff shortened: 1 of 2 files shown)


src/util/wf-autohide-window.hpp
```diff
@@ -107,6 +111,7 @@ class WayfireAutohidingWindow : public Gtk::Window
     bool input_inside_panel     = false;
     zwf_hotspot_v2 *edge_hotspot  = NULL;
     zwf_hotspot_v2 *panel_hotspot = NULL;
+    bool zwf_support = false;
```
ammen99, line 114: Do we need an additional variable for this, can't we check the output->output == NULL when we need it?

src/util/wf-autohide-window.hpp
```diff
@@ -89,11 +89,15 @@ class WayfireAutohidingWindow : public Gtk::Window
     int last_edge_offset = -1;
 
     WfOption<bool> autohide_opt;
+    bool autohide_enabled = autohide_opt;
```
ammen99, line 92: I believe we don't need this at all. We already have an autohide counter and last_autohide_value.
Looking at the code, you use this in order to disable autohiding when wayfire-shell is not enabled (right?) - you could do this check in the function which sets up autohide (which you already do) and then just print an error+return instead of connecting the autohide handlers.

ammen99 (commented):
In general, I dislike adding more variables than necessary because we end up with duplicated information which may get out of sync .. Otherwise, the idea to move each functionality to an appropriate function is good.

src/util/wf-autohide-window.hpp
lavender-agatha: Okay, agreed, I did because I was somewhat uncertain on whether the value of `output->output` could change, I thought it was something else at first sight.

Resolved in e4240e5629a2c891dd34dc9b2c17d3b810bc8e3

src/util/wf-autohide-window.hpp
lavender-agatha: @ammen99 I could, but I think this approach would fail to account for when you change the value of autohide on-the-fly (i.e., you change it to `false` and then to `true` again).

When the `autohide_opt` is updated and the callback sets up autohide again, we don't control the value of the `autohide_opt`. On the other hand, we need the `last_autohide_value` to detect whether it changed or not as well. 

I believe there must be a way to eliminate the need for the `autohide_enabled` variable, I agree that it's pretty confusing to have all those variables, I'm just not seeing it yet.

lavender-agatha:
I worked out the previous points, so I'm reopening the PR

marcusbritanicus:
@ammen99 I cam confirm that this works perfectly well all the cases I tested:
1. Autohide is off and wayfire-shell plugin in not loaded. (no autohide and exclusive zone)
2. Autohide is off and wayfire-shell plugin in loaded. (no autohide and exclusive zone)
3. Autohide is on and wayfire-shell plugin in not loaded. (no autohide and exclusive zone)
4. Autohide is on and wayfire-shell plugin in loaded. (autohide and no exclusive zone)

ammen99 (approved):
Thanks!

PR #204, opened 2023-12-30
Fix crash with parentheses in filenames

### LINKED ISSUE #203: [wf-background] crash if background image path contains parentheses
Hi,
I tried setting my background to the following:
```
[background]
cycle_timeout = 150
image = /usr/share/backgrounds/Ross_Jones_Rockpool_(Sydney)_by_Chris_Carignan.jpg
preserve_aspect = true
randomize = false
```
and got wf-background crashing (immediately and also when trying to start it manually).

I looked into this and the crash happens [here](https://github.com/WayfireWM/wf-shell/blob/f0574575bb7675740903bb0e95aefb0144929987/src/background/background.cpp#L129) -- apparently, the variable `exp` is filled with garbage.

I would say that this is actually a bug in `wordexp()`, but it is not entirely clear: [here](https://man7.org/linux/man-pages/man3/wordexp.3.html) it says that "the string must not contain [...] unescaped (, ) characters", which could imply UB, but if I interpret it correctly, the [standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wordexp.html) says that in this case "wordexp() shall fail, and the number of expanded words shall be 0" which is not the case on my system (I get we_wordc = 4). However, it would be nice to avoid crashing in this case though.

dkondor:
Fixes #203 

I used a separate assignment to avoid compiler warnings.

Note that things still won't work if the user specifies a directory with parentheses, but there will be no crash only an error message. Single image filenames with parentheses work.

### DIFF
--- a/src/background/background.cpp
+++ b/src/background/background.cpp
@@ -120,8 +120,7 @@ bool WayfireBackground::load_images_from_dir(std::string path)
     wordexp_t exp;
 
     /* Expand path */
-    wordexp(path.c_str(), &exp, 0);
-    if (!exp.we_wordv)
+    if (wordexp(path.c_str(), &exp, 0) || !exp.we_wordv)
     {
         return false;
     }
@@ -129,6 +128,7 @@ bool WayfireBackground::load_images_from_dir(std::string path)
     auto dir = opendir(exp.we_wordv[0]);
     if (!dir)
     {
+        wordfree(&exp);
         return false;
     }
 
@@ -157,6 +157,8 @@ bool WayfireBackground::load_images_from_dir(std::string path)
             }
         }
     }
+
+    wordfree(&exp);
 
     if (background_randomize && images.size())
     {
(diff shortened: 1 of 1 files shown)


dkondor:
I've changed the patch, since I've realized that `wordexp()` actually correctly returns an error code if it cannot parse its arguments. I've also added calls to `wordfree()` in the appropriate places.

src/background/background.cpp
```diff
@@ -120,15 +120,15 @@ bool WayfireBackground::load_images_from_dir(std::string path)
     wordexp_t exp;
 
     /* Expand path */
-    wordexp(path.c_str(), &exp, 0);
-    if (!exp.we_wordv)
+    if (wordexp(path.c_str(), &exp, 0) || !exp.we_wordv)
     {
         return false;
```
kode54, line 125: There should be a wordfree here if exp is nonzero, since the above condition can also fail if wordexp succeeds.

dkondor: I looked into this and actually what could happen is that if `path` is an empty string then `wordexp()` succeeds with `exp.we_wordv` being non-null (allocated properly) and `exp.we_wordc == 0`. I've modified the code to handle this case properly (now there is a separate check on `exp.we_wordc` instead).

ammen99 (approved):
Looks good, thanks!

PR #205, opened 2024-01-06
Fix setting exclusive zone when lacking zwf_shell_manager_v2 support

wb9688:


ammen99:
Isn't this almost the same as #199 ? If you can confirm #199 works I will merge that PR since it was first and also does a bit more refactoring.

marcusbritanicus:
@ammen99 I cam confirm that #199 works perfectly well all the cases I tested:
1. Autohide is off and wayfire-shell plugin in not loaded.
2. Autohide is off and wayfire-shell plugin in loaded. 
3. Autohide is on and wayfire-shell plugin in not loaded.
4. Autohide is on and wayfire-shell plugin in loaded.

ammen99:
Thanks for the PR nonetheless @wb9688 :)

PR #208, opened 2024-01-15
Listen for toggle_menu event

### LINKED ISSUE #11: Add a way to open the menu widget from the terminal
This way we can bind such a command to a keybinding and open the menu with it.

soreau:
Update wayfire-shell protocol and listen for the toggle_menu event. Fixes #11 because ipc plugin can be used to allow calling toggle_menu from a script.

Requires [wayfire PR 2097](https://github.com/WayfireWM/wayfire/pull/2097).


ammen99 (approved):
LGTM, thanks!

PR #209, opened 2024-01-16
battery: make scroll change brightness

### LINKED ISSUE #207: Add scroll on battery widget to change screen brightness
Scroll up on the battery widget -> brightness goes up
Scroll down on the battery widget -> brightness goes down

I always wanted this in stock GNOME but you had to install a third-party extension to do it.

In the meantime created a simple patch to do this, but it's not pretty. It relies on `brightnessctl`.

[patch.txt](https://github.com/WayfireWM/wf-shell/files/13930621/patch.txt)

carterprince:
`battery_scroll_sensitivity` is a new option that specifies the amount (in percentage points) that each scroll event will change the brightness. I found `1.8` to be a good default as a compromise between smooth (touchpad) scrolling and discrete (mouse-wheel) scrolling.

I wrote a method `change_brightness()` which reads and writes to `/sys/class/backlight/` to change the brightness. This requires that the running user be in the `video` group. Personally I'm fine with this but I want to hear what everyone else thinks.

If I break some conventions here I apologize; I'm not very familiar with how you're supposed to do stuff in C++. Please let me know.

Also, I haven't tested this on any machine other than my laptop. So if it doesn't work for you also let me know.

Resolves https://github.com/WayfireWM/wf-shell/issues/207

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -125,11 +125,6 @@
 		<default>32</default>
 		<min>1</min>
 	</option>
-	<option name="battery_scroll_sensitivity" type="double">
-		<_short>Battery Scroll Sensitivity</_short>
-		<default>1.8</default>
-		<min>0</min>
-	</option>
 	<option name="battery_icon_invert" type="bool">
 		<_short>Battery Icon Invert Color</_short>
 		<default>true</default>
@@ -137,6 +132,14 @@
 	<option name="battery_font" type="string">
 		<_short>Battery Font</_short>
 		<default>default</default>
+	</option>
+	<option name="battery_scrollup_command" type="string">
+		<_short>Battery Scroll Up Command</_short>
+		<default>brightnessctl -q set 10%+</default>
+	</option>
+	<option name="battery_scrolldown_command" type="string">
+		<_short>Battery Scroll Down Command</_short>
+		<default>brightnessctl -q set 10%-</default>
 	</option>
 	</group>
 	<group>
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -125,6 +125,11 @@
 		<default>32</default>
 		<min>1</min>
 	</option>
+	<option name="battery_scroll_sensitivity" type="double">
+		<_short>Battery Scroll Sensitivity</_short>
+		<default>1.8</default>
+		<min>0</min>
+	</option>
 	<option name="battery_icon_invert" type="bool">
 		<_short>Battery Icon Invert Color</_short>
 		<default>true</default>
--- a/src/panel/widgets/battery.cpp
+++ b/src/panel/widgets/battery.cpp
@@ -2,8 +2,6 @@
 #include <gtk-utils.hpp>
 #include <iostream>
 #include <algorithm>
-#include <fstream>
-#include <filesystem>
 
 #define UPOWER_NAME "org.freedesktop.UPower"
 #define DISPLAY_DEVICE "/org/freedesktop/UPower/devices/DisplayDevice"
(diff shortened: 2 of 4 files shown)


src/panel/widgets/window-list/toplevel.cpp
```diff
@@ -186,6 +186,10 @@ class WayfireToplevel::impl
         {
             menu.popup(event->button, event->time);
             return true; // It has been handled.
+        } else if ((event->type == GDK_BUTTON_PRESS) && (event->button == 2)) // middle click
```
soreau, line 189: This patch should go in a separate pull request.

carterprince: My mistake

soreau:
> I wrote a method `change_brightness()` which reads and writes to `/sys/class/backlight/` to change the brightness. This requires that the running user be in the `video` group. Personally I'm fine with this but I want to hear what everyone else thinks.

I think your original idea of using `brightnessctl` made more sense. I also think there should be an option to make the command configurable as @ammen99 suggested in #207. Maybe you can have two options, one defining the command for brighter and the other dimmer. Then you could possibly bind scroll 1-1 and eliminate the sensitivity option. Either way, you could call the options directly, as commands for the scroll events.

ammen99:
I agree with @soreau, on some of my previous laptops brightness was rather hard to change and one had to use different files in sys. Also you get the problem of min/max brightness etc. It would be better to let the user provide a command themselves. If you care about smooth scrolling too much, you could make the command a format string. I.e option value is `brightnessctl %.2f` and then the widget replaces the %.2f with the scroll delta with sprintf. Though I suspect discrete steps should be fine enough.

src/panel/widgets/battery.cpp
```diff
@@ -29,6 +38,35 @@ static std::string get_device_type_description(uint32_t type)
     return "";
 }
 
+// this works as long as the owner of the wf-panel process is in the video group
+// sudo usermod -a -G video USER
+void change_brightness(int change) {
```
marcusbritanicus, line 43: I very much doubt if this is sufficient. Please correct me if I am wrong. I think you might have a udev rule to which allows you to do this.

I am in the video group: typing groups gives my this output
```
log network scanner power users video storage optical lp audio wheel cosmos
```
and I cannot simply open `/sys/class/backlight/*/brightness` for writing.

carterprince:
Thanks for the feedback. I reverted to the original approach with the latest commit, added `battery_scrollup_command` and `battery_scrolldown_command` options, and took @ammen99's original suggestion of running commands every 3 steps of discrete scroll.

Let me know if you have any other suggestions.

metadata/panel.xml
```diff
@@ -133,6 +133,14 @@
 		<_short>Battery Font</_short>
 		<default>default</default>
 	</option>
+	<option name="battery_scrollup_command" type="string">
+		<_short>Battery Scroll Up Command</_short>
+		<default>brightnessctl -q set 10%+</default>
+	</option>
+	<option name="battery_scrolldown_command" type="string">
+		<_short>Battery Scroll Down Command</_short>
+		<default>brightnessctl -q set 10%-</default>
```
soreau, line 142: Why not do it on every discrete event at 3% by default? It seems like you're losing some granularity and possibly introducing buggy behavior by doing nothing on some scroll events.

carterprince: Sorry, I was just doing what I thought ammen99 originally suggested. It just made sense to me because running so many commands at once makes the brightness update laggy on my system.

ammen99: Sounds like another thing which ought to be an option (granularity)

src/panel/widgets/battery.cpp
```diff
@@ -174,6 +174,23 @@ void WayfireBatteryInfo::update_details()
     }
 }
 
+void WayfireBatteryInfo::on_battery_scroll(GdkEventScroll *event)
+{
+    if (scroll_counter <= 0) {
```
ammen99, line 179: How does GDK behave when we have smooth scroll, do we get separate events for discrete and smooth scrolling, or are they bunched here together? Can `delta_*` be a floating point number?

In general I am kinda hesitant to have a hard threshold like this (i.e update every 3 events). Wouldn't it be better to update on every X units of scroll, or even better, every <N> milliseconds, where N is a configurable period of time?

soreau: To complicate things further, there is also GdkEventScroll.[GdkScrollDirection](https://developer-old.gnome.org/gdk3/stable/gdk3-Event-Structures.html#GdkScrollDirection), which tells when the delta_x/y values are valid to use or which direction to scroll.

AKArien:
Closing as #327 gives backlight level control in it’s own widget. Hopefully, it is of your taste !

PR #210, opened 2024-01-16
panel: middle click to close window

carterprince:


src/panel/widgets/window-list/toplevel.cpp
```diff
@@ -186,6 +186,10 @@ class WayfireToplevel::impl
         {
             menu.popup(event->button, event->time);
             return true; // It has been handled.
+        } else if ((event->type == GDK_BUTTON_PRESS) && (event->button == 2)) // middle click
```
soreau, line 189: This probably should only be enabled with an option that is disabled by default.

carterprince: Done!

metadata/panel.xml
```diff
@@ -74,6 +74,10 @@
 		<_short>Background Color</_short>
 		<default>gtk_headerbar</default>
 	</option>
+	<option name="middle_click_close" type="bool">
```
soreau, line 77: This should be taken out of the `General` group and placed in a new `Window List` group.

ammen99 (approved):
LGTM, thanks!

PR #214, opened 2024-01-23
window-list: Allow smaller panel heights

soreau:
Set the scrolled window policy to NEVER instead of AUTOMATIC to allow it to become smaller than 32 pixels in height.

### DIFF
--- a/src/panel/widgets/window-list/toplevel.cpp
+++ b/src/panel/widgets/window-list/toplevel.cpp
@@ -298,9 +298,10 @@ class WayfireToplevel::impl
 
     void set_app_id(std::string app_id)
     {
+        WfOption<int> minimal_panel_height{"panel/minimal_height"};
         this->app_id = app_id;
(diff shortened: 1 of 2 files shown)


mrsteve0924:
trying to build from window-list-height branch and getting  errors.  not sure why it says "wayfire found no".  and i have glm installed. 

Found pkg-config: YES (/usr/bin/pkg-config) 2.1.0
Found CMake: /usr/bin/cmake (3.28.2)
Run-time dependency wayfire found: NO (tried cmake)

meson.build:16:17: ERROR: Dependency lookup for wayfire with method 'pkgconfig' failed: Could not generate cflags for wayfire:
Package glm was not found in the pkg-config search path.
Perhaps you should add the directory containing `glm.pc'
to the PKG_CONFIG_PATH environment variable
Package 'glm', required by 'wf-config', not found

ammen99 (approved):
Thanks :)

PR #225, opened 2024-03-27
window-list: use ~/.local/share when looking for apps / icons

lcolitti:
The icons in the window list are loaded by parsing the desktop files for a few variations of the window's app ID in a few known locations.

Add /.local/share/applications to the list of locations it looks in, to ensure that apps installed in the user's home directory are considered as well.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,7 +5,7 @@ jobs:
   test_musl_gcc:
     name: "Test with with GCC/musl/libstdc++/BFD on Alpine Linux"
     runs-on: ubuntu-latest
-    container: alpine:edge
+    container: alpine:latest
     steps:
     - run: echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories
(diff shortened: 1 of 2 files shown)


soreau:
Somehow I'm thinking this should rely on XDG_DATA_DIRS to do the right thing. Consider Gio::DesktopAppInfo::create() [here](https://developer-old.gnome.org/glibmm/stable/classGio_1_1DesktopAppInfo.html).

ammen99:
I think @soreau is right, probably if we try the create() method first, we would most likely be able to cover this case and even a few others.

lcolitti:
> Somehow I'm thinking this should rely on XDG_DATA_DIRS to do the right thing. Consider Gio::DesktopAppInfo::create() [here](https://developer-old.gnome.org/glibmm/stable/classGio_1_1DesktopAppInfo.html).

Ah yes, that looks much better. Thanks for the suggestion!

lcolitti:
Looks like the CI tests are broken for unrelated reasons:

```
  Error relocating /usr/lib/libcurl.so.4: SSL_get0_group_name: symbol not found
  Error: The process '/usr/bin/git' failed with exit code 128
```

ammen99:
@lcolitti You can probably use `alpine:latest` instead of `alpine:edge`

lcolitti:
> @lcolitti We can probably use `alpine:latest` instead of `alpine:edge` to make the CI work again

Yup, definitely looks happier after changing that.

soreau:
Bonus points for removing more lines than added. 👍

PR #226, opened 2024-04-02
Add "Fill and Crop" option for wf-background. Pt 2

trigg:
I felt strongly about https://github.com/WayfireWM/wf-shell/pull/58 so I replicated the fork against master and followed the requested changes.

After some testing with a variety of images and screens I made further edits:
- Added an `image_scale` value that gets passed into the render logic for each image
- Added cairo context save/restore and placed a scale inside it to allow the image to be scaled up in-place without loading it a second time.

These were necessary as loading tall images in ultrawide screens or visa-versa caused a crash from attempting to create a Pixbuf that was too large.

I have attempted to stick to the coding style of the project but feel free to correct me on it.

### DIFF
--- a/metadata/background.xml.in
+++ b/metadata/background.xml.in
@@ -21,9 +21,21 @@
 		<_short>Randomize</_short>
 		<default>false</default>
 	</option>
-	<option name="preserve_aspect" type="bool">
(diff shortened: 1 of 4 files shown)


ammen99 (approved):
Tested and seems to work, thanks!

PR #230, opened 2024-04-11
Collection of Menu alterations

trigg:
Options added:
- Application icons in list form `panel/menu_list`
- Menu shows application categories `panel/menu_show_categories`
- Categories width `panel/menu_min_category_width`

New Menu Features:
- Categories to the left of application list
- Keyboard navigation of applications & categories
- Right click an application to choose between alternative actions
- Changing search term selects first item in list
- Categories hidden if no applications exist for it

Commits include work done for a different PR merged here and then reverted as the patch was unrelated.

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -20,9 +20,7 @@ const std::string default_icon = ICONDIR "/wayfire.png";
 
 WfMenuCategoryDefinition::WfMenuCategoryDefinition(std::string _name, std::string _icon_name):
     name(_name), icon_name(_icon_name)
-{
-
-}
+{}
 
 std::string WfMenuCategoryDefinition::get_name()
 {
@@ -34,7 +32,8 @@ std::string WfMenuCategoryDefinition::get_icon_name()
     return icon_name;
 }
 
-WfMenuCategoryButton::WfMenuCategoryButton(WayfireMenu *_menu, std::string _category, std::string _label, std::string _icon_name) :
+WfMenuCategoryButton::WfMenuCategoryButton(WayfireMenu *_menu, std::string _category, std::string _label,
+    std::string _icon_name) :
     Gtk::Button(), menu(_menu), category(_category), label(_label), icon_name(_icon_name)
 {
     m_image.set_from_icon_name(icon_name,
@@ -87,10 +86,10 @@ WfMenuMenuItem::WfMenuMenuItem(WayfireMenu *_menu, Glib::RefPtr<Gio::DesktopAppI
             m_padding_box.pack_start(*action_button, false, false);
 
             action_button->signal_clicked().connect(
-                [this] () {
-                    m_action_menu.popup_at_widget(this, Gdk::GRAVITY_NORTH_EAST, Gdk::GRAVITY_NORTH_WEST, NULL);
-                }
-            );
+                [this] ()
+            {
+                m_action_menu.popup_at_widget(this, Gdk::GRAVITY_NORTH_EAST, Gdk::GRAVITY_NORTH_WEST, NULL);
+            });
         }
     } else
     {
@@ -109,19 +108,20 @@ WfMenuMenuItem::WfMenuMenuItem(WayfireMenu *_menu, Glib::RefPtr<Gio::DesktopAppI
         m_padding_box.pack_start(m_right_pad);
     }
 
-    for(auto action : app->list_actions())
+    for (auto action : app->list_actions())
     {
         auto menu_item = new Gtk::MenuItem();
         menu_item->set_label(m_app_info->get_action_name(action));
         menu_item->show();
         menu_item->signal_activate().connect(
-            [this, action] () {
-                m_app_info->launch_action(action);
-                menu->hide_menu();
-            }
-        );
+            [this, action] () 
+        {
+            m_app_info->launch_action(action);
+            menu->hide_menu();
+        });
         m_action_menu.append(*menu_item);
     }
+
     add(m_padding_box);
     /* Can accept Enter/Return from search_box */
     set_can_default(true);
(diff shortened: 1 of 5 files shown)


src/panel/widgets/menu.cpp
```diff
@@ -148,9 +253,74 @@ void WayfireMenu::load_menu_item(AppInfo app_info)
 
     loaded_apps.insert({name, exec});
 
-    items.push_back(std::unique_ptr<WfMenuMenuItem>(
-        new WfMenuMenuItem(this, app_info)));
-    flowbox.add(*items.back());
+    /* Check if this has a 'OnlyShownIn' for a different desktop env 
+       If so, we throw it in a pile at the bottom just to be safe */
+    if (!desktop_app_info->get_show_in("wayfire")){
```
trigg, line 258: This filters out configuration programs for alternative window managers/desktops

I'm not happy about the hardcoded "wayfire" as it claims it will read from `XDG_CURRENT_DESKTOP` if given null. That only buys me crashes when I try.

src/panel/widgets/menu.hpp
```diff
@@ -15,10 +15,39 @@
 class WayfireMenu;
 using AppInfo = Glib::RefPtr<Gio::AppInfo>;
 
-class WfMenuMenuItem : public Gtk::HBox
+class WfMenuCategoryDefinition 
 {
```
trigg, line 19: This is probably over engineered for the purpose, but keeping two string-to-string hashmaps felt a bit more open to error

src/panel/widgets/menu.cpp
ammen99: maybe XDG_CURRENT_DESKTOP is not set? Wayfire doesn't set it automatically, the user is supposed to do that manually.

src/panel/widgets/menu.cpp
```diff
@@ -494,11 +712,31 @@ static void app_info_changed(GAppInfoMonitor *gappinfomonitor, gpointer user_dat
...
+    category_definitions["Development"] = std::make_unique<WfMenuCategoryDefinition>("Development", "applications-development");
+    category_definitions["Graphics"] = std::make_unique<WfMenuCategoryDefinition>("Graphics", "applications-graphics");
+    category_definitions["AudioVideo"] = std::make_unique<WfMenuCategoryDefinition>("Multimedia", "applications-multimedia");
+    category_definitions["Game"] = std::make_unique<WfMenuCategoryDefinition>("Games", "applications-games");
+    category_definitions["Science"] = std::make_unique<WfMenuCategoryDefinition>("Science", "applications-science");
+    category_definitions["Settings"] = std::make_unique<WfMenuCategoryDefinition>("Settings", "preferences-desktop");
+    category_definitions["System"] = std::make_unique<WfMenuCategoryDefinition>("System", "applications-system");
+    category_definitions["Utility"] = std::make_unique<WfMenuCategoryDefinition>("Accessories", "applications-utilities");
+    category_definitions["Hidden"] = std::make_unique<WfMenuCategoryDefinition>("Other Desktops", "user-desktop");
+
```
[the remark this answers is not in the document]
ammen99: I do think it makes sense to have one fallback where everything not fitting other categories is present. I'm certain some apps will be mis-labeled, so this way the user can still see them ..

src/panel/widgets/menu.hpp
ammen99: It's fine :)

src/panel/widgets/menu.hpp
```diff
@@ -111,16 +143,22 @@ class WayfireMenu : public WayfireWidget
     void on_search_changed();
     void on_popover_shown();
 
-    std::vector<std::unique_ptr<WfMenuMenuItem>> items;
     /* loaded_apps is a list of the already-opened applications + their execs,
      * so that we don't show duplicate entries */
     std::set<std::pair<std::string, std::string>> loaded_apps;
+    std::unordered_map<std::string, std::vector<Glib::RefPtr<Gio::DesktopAppInfo>>> items;
```
ammen99, line 159: It seems that we're still duplicating the list of categories a few times (we have map name -> items, name -> definition, also a list of orders). Since you already have a struct for CategoryDefinitions, I think that we can at least merge the `items` map and make the list of AppInfos part of the `WfMenuCategoryDefinition` struct.

The order seems a bit more difficult so it could remain separate I'd say.

About the order,

src/panel/widgets/menu.cpp
trigg: I can assure you it is!

I'll do more testing on this, as it seems unlikely that it doesn't do what the documentation claims it does

src/panel/widgets/menu.cpp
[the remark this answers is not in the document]
trigg: Prefer safe to sorry, I see that.

Then potentially a bool to hide the category?

src/panel/widgets/menu.hpp
trigg: Solid point, the potential of this evaded me

src/panel/widgets/menu.cpp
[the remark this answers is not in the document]
ammen99: I don't think we should hide any applications from the user, then again, I usually don't use the menu, so if you feel strongly in either direction, I'd trust your intuition.

ammen99 (approved):
In general LGTM, let me know if you want to make any more changes, otherwise, I will merge.

Also a note, I usually squash all commits together to avoid style fix commits in the history, if you want your work to be remain in multiple commits, please squash manually. Otherwise I'll just squash and you don't have to do anything.

trigg:
> ... otherwise, I will merge.

That's fine. While I'm unsure everyone will agree with me about the defaults, they're certainly acceptable defaults.

> ... Otherwise I'll just squash and you don't have to do anything.

Go for it, I'm not precious about it.

ammen99:
Alright, I think even if someone disagrees, we can add more options in follow-up PRs, you did the harder part ;) Thanks!

PR #231, opened 2024-04-15
menu: Allow sorting by relevance

trigg:
When filtering/searching the usual text sort should be switched to a relevance sort.


ammen99 (approved):
LGTM, thanks!

Issue #211, opened 2024-01-17
[Feature] Add panel separator as a widget


mrsteve0924:
I use xfce and they have a panel plugin to add a visual separator. 

I kind of hacked one together using launcher_cmd and launcher_icon in wf-shell.ini.  But it's limited to where I can place it.

> launcher_cmd_separator2 = 
launcher_icon_separator2 = /usr/share/icons/Papirus/symbolic/apps/separator-symbolic.svg
![wf-panel](https://github.com/WayfireWM/wf-shell/assets/37476191/eb49a09f-8c8a-48e5-9c57-1ae502f0c6f4)

here's how the separators look on my xfce panel
![xfce-panel](https://github.com/WayfireWM/wf-shell/assets/37476191/ab108645-2e59-4f33-aab0-86690f2df833)

https://docs.xfce.org/xfce/xfce4-panel/separator

ammen99:
fixed in #233, you can use `separator<width in pixels>` just like `spacing` in the config file.

PR #233, opened 2024-04-16
Visible separator

trigg:
In answer to #211 

Allow for rendering a single thin line using primary color. Should be configurable from CSS and by default matchs system theme.


ammen99:
Wouldn't it be better to make it so that we support both `spacingN` and `separatorN` as widgets, so that the user can select which one to use where?

src/panel/panel.cpp
```diff
@@ -230,6 +231,20 @@ class WayfirePanel::impl
...
         }
+        std::string separator = "separator";
+        if (name.find(separator) == 0)
+        {
+            auto pixel_str = name.substr(separator.size());
+            int pixel = std::atoi(pixel_str.c_str());
+
+            if (pixel <= 0)
+            {
+                std::cerr << "Invalid separator, " << pixel << std::endl;
```
ammen99, line 243: Might be better to print `pixel_str` so that the user sees which string is problematic

ammen99 (approved):
LGTM, one small nit and we can merge.

src/panel/panel.cpp
trigg: Shall I do the same to spacing while I'm here?

ammen99: That actually reminds me, can't we extract it into a function?

trigg: Is this implementation what you had in mind?

mrsteve0924:
nice! cant wait to check it out.

ammen99 (approved):
Thanks!

PR #234, opened 2024-04-16
window-list: Dragging files to window list and hovering activates window

### LINKED ISSUE #202: Dragging files between programs using the window list not working
This is a common feature available on other panels and on Windows.
TL;DR it seems that dragging a file from my file manager to another program, by holding the file over that other program's running icon on the window list, does not work.
Pinging @soreau for advice.

trigg:
Intended to fix #202

### DIFF
--- a/src/panel/widgets/window-list/toplevel.cpp
+++ b/src/panel/widgets/window-list/toplevel.cpp
@@ -480,6 +480,10 @@ class WayfireToplevel::impl
 
     ~impl()
     {
+        if (m_drag_timeout)
+        {
+            m_drag_timeout.disconnect();
+        }
         zwlr_foreign_toplevel_handle_v1_destroy(handle);
     }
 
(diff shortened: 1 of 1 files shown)


src/panel/widgets/window-list/toplevel.cpp
```diff
@@ -40,6 +41,7 @@ class WayfireToplevel::impl
     Gtk::Menu menu;
     Gtk::MenuItem minimize, maximize, close;
     Glib::RefPtr<Gtk::GestureDrag> drag_gesture;
+    sigc::connection m_drag_timeout;
```
ammen99, line 44: Should we disconnect the connection in the object destructor as well? Or are we somehow guaranteed we'll get a drag leave first?

trigg: Most likely will but actually it's probably better to be safe.

ammen99 (approved):
Thanks!

PR #237, opened 2024-04-18
Lauchers fixes

trigg:
launchers: store one pixbuf and render at needed scale to avoid repeatedly loading from disk
launchers: remove widget resize from animate to avoid recalculating all panel widgets
launchers: `.desktop` launchers should fall back to `image-missing` icon
launchers: cmd launchers may use themed icons
launchers: change of config should take without reload

### DIFF
--- a/src/panel/widgets/launchers.cpp
+++ b/src/panel/widgets/launchers.cpp
@@ -3,6 +3,7 @@
 #include <glibmm/spawn.h>
 #include <gdkmm/pixbuf.h>
 #include <gdkmm/general.h>
+#include <gdkmm/general.h>
 #include <gtkmm/icontheme.h>
 #include <gdk/gdkcairo.h>
 #include <cassert>
@@ -52,11 +53,11 @@ struct DesktopLauncherInfo : public LauncherInfo
         {
             std::cerr << "Failed to load icon \"" << icon << "\"" << std::endl;
             return theme->load_icon("image-missing", size)
-               ->scale_simple(size, size, Gdk::INTERP_BILINEAR);
+                ->scale_simple(size, size, Gdk::INTERP_BILINEAR);
         }
 
         return theme->load_icon(icon, size)
-               ->scale_simple(size, size, Gdk::INTERP_BILINEAR);
+            ->scale_simple(size, size, Gdk::INTERP_BILINEAR);
     }
 
     std::string get_text()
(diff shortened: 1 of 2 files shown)


src/panel/widgets/launchers.cpp
```diff
@@ -149,26 +155,56 @@ bool WfLauncherButton::initialize(std::string name, std::string icon, std::strin
...
+            {
+                evbox.queue_draw();
+            }
+        }
+
+        return false;
     });
 
-    current_size.set(base_size, base_size);
-    evbox.property_scale_factor().signal_changed()
```
ammen99, line 165: We still need to handle the scale change event, because the icon size changes (for example if we render with scale=2.0 and we want the icons to have size 50px, those 50px are logical sizes and we actually need size 100px)

src/panel/widgets/launchers.cpp
```diff
@@ -149,26 +155,56 @@ bool WfLauncherButton::initialize(std::string name, std::string icon, std::strin
...
+    update_size();
 
     evbox.set_tooltip_text(info->get_text());
     return true;
 }
 
+void WfLauncherButton::update_size()
+{
+    base_size = size / LAUNCHERS_ICON_SCALE;
+    image     = info->get_pixbuf(size);
```
ammen99, line 202: The size of the icon we load should be multiplied with the device scale factor.

ammen99 (changes requested):

src/panel/widgets/launchers.cpp
```diff
@@ -2,6 +2,8 @@
 #include <giomm/file.h>
 #include <glibmm/spawn.h>
 #include <gdkmm/pixbuf.h>
+#include <gdkmm/general.h>
+#include <gdkmm/general.h>
```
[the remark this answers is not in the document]
trigg: I must've introduced a mistake trying to use uncrustify on local machine. I'll splat that now

src/panel/widgets/launchers.cpp
trigg: I've done this, the variables held are kept in widget space as that's what Cairo expects us to use. It will now only account for scale on image load, to ensure we don't get anti-aliased images.

trigg:
Closing in advance of a better plan

PR #238, opened 2024-04-18
Allow smaller window list

### LINKED ISSUE #213: make window-list height resizeable
window-list seems to hold the panel height at a very high number of pixels, only way to get the panel size down currently is to remove the window-list. please add a config option to set the window-list height

trigg:
window-list: add font option
window-list: add reduced padding/small option

Fixes #213

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -356,6 +356,14 @@ Set to -1 to only run it by clicking the button.
 		<_short>Middle Click Closes Windows</_short>
 		<default>false</default>
 	</option>
+	<option name="window_list_font" type="string">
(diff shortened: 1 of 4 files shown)


ammen99 (commented):
At this point I'm starting to wonder, shouldn't we actually work to make wf-panel more easily configurable via a css file? We have some support for that but I suppose adding classes/IDs to all kinds of widgets will make this much more flexible and will remove the need for so many options, no?

trigg:
Potentially yes.

If the documentation is lacking or gets put out of date it would be worth a lot less than all these options to most users.

PR #239, opened 2024-04-20
panel: add css classes to widgets

trigg:
The thin end of the CSS wedge.

Adds Classes where appropriate
Removes state flags from volume control
Replaces hardcoded wayfire icon with theme-search icon.


ammen99 (approved):
Thanks!

PR #243, opened 2024-04-21
launcher: remove features

trigg:
launcher: remove zoom animation
launcher: use in-memory appinfo if needed
menu: fix an assumption on launcher
menu: use gtk util icons


ammen99:
There is somethign here which doesn't work properly, before, one could set full shell commands but I don't think this works anymore because you wrap everything in a DesktopAppInfo (I think at least that's the reason it fails)

ammen99:
Something like https://0x0.st/XH1j.diff will be needed

ammen99:
The metadata in `metadata/panel.xml` needs to also be cleaned up to remove the obsolete options (at least the animation duration, idk whether there are others as well).

ammen99:
Thanks!

PR #244, opened 2024-04-21
gtk-utils: set icon changes

trigg:
gtk-utils: set_image_icon use both theme and filesystem
gtk-utils: set_image_icon stores at double necessary scale to allow unblurred scaling


trigg:
Moved the filesystem & theme icon logic away up to here, since we're not doing slight variations of this in some widgets now

PR #245, opened 2024-04-21
tray: allow css styling

trigg:
tray: add icons to button
tray: fix click logic for button
tray: attach popup menu to button

### DIFF
--- a/src/panel/widgets/tray/item.cpp
+++ b/src/panel/widgets/tray/item.cpp
@@ -70,7 +70,7 @@ void StatusNotifierItem::init_widget()
     setup_tooltip();
     init_menu();
     auto style = get_style_context();
-    style->add_class("tray-box");
+    style->add_class("tray-button");
     style->add_class("flat");
 
     signal_clicked().connect([this] () -> void
--- a/src/panel/widgets/tray/item.cpp
+++ b/src/panel/widgets/tray/item.cpp
@@ -98,6 +98,7 @@ void StatusNotifierItem::init_widget()
         {
             return true;
         }
+
         const auto ev_coords = Glib::Variant<std::tuple<int, int>>::create({ev->x, ev->y});
         const guint menu_btn = menu_on_middle_click ? GDK_BUTTON_MIDDLE : GDK_BUTTON_SECONDARY;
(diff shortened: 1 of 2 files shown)


trigg:
Switch the eventbox which is not usable via CSS to a button, and account for the mouse-press differences.

Attach popup corner to icon corner, Gtk accounts for orientation issues automatically

NamorNiradnug:
Hi, thanks for your PR!

As far as I remember GtkButton doesn't handle scroll events while EventBox does, and that's the reason EventBox is used currently. Please consider applying some workaround to keep scrolling above tray items working.

trigg:
I can't say I know a single system item that uses scroll wheel, got a program I can test on?

trigg:
From my tests the widget acts the same as before for scroll, but everything I have runs through the GDK_SCROLL_SMOOTH code path... even the discreet scroll wheel.

soreau:
> From my tests the widget acts the same as before for scroll, but everything I have runs through the GDK_SCROLL_SMOOTH code path... even the discreet scroll wheel.

If you read [the documentation](https://docs.gtk.org/gdk3/struct.EventScroll.html), you must first consider the value of the `direction` field to see if the event is GDK_SCROLL_[CARDINAL] or GDK_SCROLL_SMOOTH. If it is the latter, only then can you use the `delta` fields.

trigg:
Allow me to rephrase: 

All my devices report an ev->direction value of GDK_SCROLL_EVENT. 

I am unable to do testing on the alternative codepaths for GDK_SCOLL_* but can't see anything there that would fail when using Gtk::Button in place of Gtk::EventBox.

trigg:
With a bunch of `std::cout` calls plastered about I've come to a conclusion:

On (at least) my version of GTK Every Single scroll event will pose as Smooth scroll until I remove the flag here.

https://github.com/WayfireWM/wf-shell/blob/7320d4046a2bb95c5a9649929d5b1a51c112a33a/src/panel/widgets/tray/item.cpp#L122

After this is removed I get `GDK_SCROLL_UP` and `GDK_SCROLL_DOWN` on my wheel correctly, and now every 5th click will actually send the signal to the tray program.

So this rules out the previous statement that I couldn't test that code path. I just needed to jump a few more hoops to reach it.

This branch doesn't alter scroll wheel responsiveness on any of my tests.

PR #246, opened 2024-04-21
window-list: attach popup menu to button

trigg:


trigg:
Aligns edge of popup to edge of the button, orientation is corrected by Gtk by default if top/bottom is selected

ammen99:
So this changes the position of the menu popup, a question at people who use this menu (@soreau? and probably others) which one do you prefer:

- New layout:
slurped

- Old layout:
slurped


I.e should the menu be centered on where the user clicked (which is somewhere in the middle of the button) or below the button.

trigg:
And one with a bit more space on the list

trigg:
For reference, I feel this is correct as it matches clock, menu & (more recently) tray.

ammen99:
Thanks!

PR #249, opened 2024-04-22
background: free old image after transition

trigg:


ammen99 (approved):
LGTM, thanks! Does this help with reduction of RAM usage? (I expect it should halve it)

trigg:
This PR lowered the ram usage in my tests, but unfortunately not by half.

ammen99:
> This PR lowered the ram usage in my tests, but unfortunately not by half.

I guess it also depends on the background image's resolution. But any reduction is good :)

PR #251, opened 2024-04-27
window-list: cpu usage fix

### LINKED ISSUE #250: window-list high cpu usage when full
window list gets into a loop when too many windows are open

the console will log similar to:
```
target button 1242
set width 28
```

on repeat until the number of windows comes down. This brings CPU usage up to about 40% in my tests

trigg:
Quick check to avoid recalculating widget size on a loop when the window-list is over-full.

fixes #250

### DIFF
--- a/src/panel/widgets/window-list/toplevel.cpp
+++ b/src/panel/widgets/window-list/toplevel.cpp
@@ -397,6 +397,7 @@ class WayfireToplevel::impl
         {
             return;
         }
+
         this->max_width = width;
         if (max_width == 0)
         {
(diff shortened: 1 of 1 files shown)


src/panel/widgets/window-list/toplevel.cpp
```diff
@@ -359,8 +359,12 @@ class WayfireToplevel::impl
     {
         this->title = title;
         button.set_tooltip_text(title);
+        label.set_text(title);
 
-        set_max_width(max_width);
+        /* Force a refresh for new title contents */
```
ammen99, line 364: instead of hacking this, can't we just split set_max_width into the actual setter and another function which just reloads the contents after a change?

trigg: Done that now

PR #255, opened 2024-04-29
Move some more style code into CSS & misc extra CSS work

trigg:
panel: prioritise default.css slightly lower than user-supplied css
panel: remove static-value set-spacing and move to default.css
notifications: add css classes
command-output: add class defining icon-direction to allow css themes to use this


ammen99 (approved):
Thanks!

Issue #266, opened 2024-07-20
Tray tooltip variables are used without checking for escape sequences


trigg:
https://github.com/WayfireWM/wf-shell/blob/3e7576f1a74abf746d3b6c3da914a428bbbd2ca7/src/panel/widgets/tray/item.cpp#L188

both `tooltip_title` and `tooltip_text` should be checked for escapable characters before passing to `set_markup`

The equivalent python code I've written previously to clean this up is

```
def sanitize_string(self, string):
        """Sanitize a text message so that it doesn't interfere with Pango's XML format"""
        string = string.replace("&", "&amp;")
        string = string.replace("<", "&lt;")
        string = string .replace(">", "&gt;")
        string = string.replace("'", "&#39;")
        string = string.replace("\"", "&#34;")
        return string
```

And the error can be triggered with this short python code

```
import gi
gi.require_version("Gtk", "3.0")
gi.require_version('AppIndicator3', '0.1')
# pylint: disable=wrong-import-position,wrong-import-order
from gi.repository import Gtk, Gdk, GLib  # nopep8
from gi.repository import AppIndicator3

window = Gtk.Window()
window.set_title("& WHY")
window.show()
menu = Gtk.Menu()
ind = AppIndicator3.Indicator.new(
    "test",
    "test",
    AppIndicator3.IndicatorCategory.APPLICATION_STATUS)
ind.set_title("& discover")
ind.set_status(AppIndicator3.IndicatorStatus.ACTIVE)
ind.set_menu(menu)
window.connect('destroy', Gtk.main_quit)
Gtk.main()

```

NamorNiradnug:
Hi! This is actually not a bug but a feature because `StatusNotifierItem` allows markup to tags in tooltip: https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/Markup/

PR #268, opened 2024-07-30
panel: sanitize variables added to markup

trigg:
I've intentionally side-stepped touching `command-output` as it is fully possible that some users are intentionally writing markup out from a command to the panel.


trigg:
Fixes #266

NamorNiradnug:
Hi! Thanks for your PR.
Current behavior is actually not a bug but a feature because `StatusNotifierItem` protocol allows markup tags in tooltip text: https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/Markup/

trigg:
Interesting, that is an issue.

This came to light because an app (Spotify i believe) was using a loose ampersand which was causing a warning and incorrect label.

NamorNiradnug:
What if we apply `escape_text` on summary text only, but not on tooltip?

trigg:
Looking back at the initial bug, yes that would cover it. Might still be scope for the bug to creep back in as the documentation you linked doesn't mention escaping & at all.

NamorNiradnug:
> Looking back at the initial bug, yes that would cover it. Might still be scope for the bug to creep back in as the documentation you linked doesn't mention escaping & at all.

I think that's a good solution then because it's simple and the specification doesn't mean markup for title. Escaping `&` and other symbols manually would make the code complicated and I don't think it is worth it.

NamorNiradnug:
> @trigg @NamorNiradnug what should be done with this PR? I have no idea about the notification widget at this point, so I trust you can figure it out between yourselves ;) 

Current solution degrades the support of markup. Hence I think we shouldn't merge it yet.

trigg:
While I theoretically agree, we're using it to stitch together unknown-source strings into existing markup mostly.

If it's the only thing going into set markup we can probably do the `if` and not use the `markup_escape` helper in that case

PR #272, opened 2024-09-17
meson.build: Add missing glm dependency check

kode54:
Add a dependency check for glm, as it's now required to build.


kode54:
This is a sign that Wayfire-git now depends on glm as a system dependency, not as a makedepend.

PR #273, opened 2024-09-21
wf-dock: improve finding icon for some GNOME apps

dkondor:
This is a much more limited attempt than #82 

Running on Ubuntu 24.04, this fixes finding icons for at least gedit (org.gnome.gedit), evince (org.gnome.Evince) and devhelp (org.gnome.Devhelp).

Let me know if moving this to a common util (with wf-panel) is still desired.


dkondor:
Closing this, as I realized I used my master branch for this PR and I need to update it. I'll leave #274 open and let me know there if this limited  approach is preferred.

PR #274, opened 2024-09-21
wf-dock: try to find icons for snap apps

dkondor:
Depends on #273 but this includes a bit more controversial heuristic, using the format used by (some / most?) snap apps.

Tested on Ubuntu 24.04 with firefox and thunderbird installed only via snap:
```
/var/lib/snapd/desktop/applications/firefox_firefox.desktop
/var/lib/snapd/desktop/applications/thunderbird_thunderbird.desktop
```


ammen99 (approved):
Thanks!

Issue #280, opened 2024-11-30
Invalid panel background color in config file


vdegenne:
I'm trying to change the panel background color,
I copied the value from  `wf-shell.ini.example` to make sure I'm not doing anything wrong (`background_color = $555555`).
But then upon wf-panel restart I have this error in the logs:

"Invalid panel background color in config file"

Thanks for your help.

soreau:
wf-shell.ini.example is wrong in this case, it should be `\#` instead of `$`.

vdegenne:
Thanks!

Btw I have a very weird output from `wf-panel`, I don't know if I should ignore them, but for now it works.

```
wf-panel
II 30-11-24 12:41:16.220 - [src/file.cpp:579] Reading XML configuration options from directory /usr/share/wayfire/metadata/wf-shell
II 30-11-24 12:41:16.220 - [src/file.cpp:529] Reading XML configuration options from file /usr/share/wayfire/metadata/wf-shell/background.xml
II 30-11-24 12:41:16.220 - [src/file.cpp:529] Reading XML configuration options from file /usr/share/wayfire/metadata/wf-shell/dock.xml
II 30-11-24 12:41:16.220 - [src/file.cpp:529] Reading XML configuration options from file /usr/share/wayfire/metadata/wf-shell/panel.xml
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
Error loading file:
```

soreau:
Please create a new issue for new questions. Leaving this one open until we fix the example wf-shell config file.

vdegenne:
https://github.com/WayfireWM/wf-shell/pull/281

PR #281, opened 2024-11-30
fix(example): show proper way to change panel background color

vdegenne:
Fixes https://github.com/WayfireWM/wf-shell/issues/280


soreau:
Looks good to me!

For what it's worth, you can use `Fixes #<issue_number>.` in the commit message or the description, so that it automatically references the issue number and closes it once this is merged.

ammen99 (approved):
thanks :)

PR #285, opened 2025-01-11
center multiline clock text

### LINKED ISSUE #282: Newline for clock
I prefer the time above the date, so if we could put something like \n in the feild to do this, that would be much appreciated.

NamorNiradnug:
Fixes #282


ammen99 (approved):
thanks!

Issue #254, opened 2024-04-28
default.css will not update to newer contents automatically


trigg:
I'm torn as to what I think is the correct path here.

# Currently

- default.css is populated on first run
- if edited by user, it is not overwritten
- if deleted by user, it is not replaced
- if edited upstream, new edits in master/stable do not propagate to user

# Worse idea

- each update default.css is replaced with a new copy
- if edited by user, it magically reverts
- if deleted by user, it magically reappears
- if edited upstream, new content is automatically applied

# Midway solution

- instead of a copy, use a symlink to the root-owned file
- user will not be able to edit the contents
- if deleted by user, it is not replaced
- if edited upstream, new content is automatically applied

My personal opinion is to switch over to the midway solution as it still allows user to have ultimate say with the lowest likelihood of breakage from updates.

Naturally this is a call for opinions on the matter.

trigg:
This has been dealt with. We settled on loading defaults from resources, not user dir

PR #290, opened 2025-02-16
Ported to GTK4

trigg:
Extensive changes to bring the code base up to working on GTK4.

Dependencies:
- gtkmm 3 > 4 (same for gdk, glib so on)
- gtk-layer-shell > gtk4-layer-shell
- libdbusmenu-gtk3 > libdbusmenu-glib

Material changes overview:
- The menus on system tray icons had to be rewritten from scratch as the supporting library appears to not be available for gtk4
- - Menu & items are regenerated when changed over the dbus wire
- - Icons appear to be getting ignored by Popover Menu. From all my reading this is intentional on GTK4s part
- - Dbus library answers that there are no tickbox or radio button variants despite testing intentionally for them. These menu items function correctly but don't display the radio/checkmark
- - I am convinced there are some memory leaks here and am squashing them as I find them
- The entirety of the event handling side of GTK has been rewritten and redefined.
- - System tray icons, Window list windows and application menu context buttons are all now MenuButton allowing us to connect menu options to them
- - EventBox no longer exists, so event callbacks are connected in directly to widgets
- - Volume control slider popup stops scroll wheel & swipe events on button, so the popup no longer shows when volume changes
- Size allocation events, Redraw events, Scale change events have been removed from gtk4 and considered "the wrong way to do things" for most uses
- - Widgets should not base their internal logic on position & size
- - Scale is left to GTK4 to account for in all cases
- - Icon size should be changed with CSS rules only
- - All `icon_size` options have been fed into CSS Styles that are generated & updated from config
- Output management and hotplugging is managed differently and worth double and triple checking by people with less boring monitor layouts than mine

While I was in the area:
- Implemented CSS directory loading for dock and background
- Added some hover, open window & close window animations to default css
- Added a dock window-close-animation option as it has adverse effects if you enable it but don't update your default css
- Realised we'd be better served with a Gtk Revealer... but that's a bigger task that I'll do in a separate attempt. 
- Reworked Background, needs sanity check as I'm sure it'll leak memory

I'm sure there's more, and will edit this as and when possible.

I look forward to uncrustify.


.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "subprojects/wayland-logout"]
 	path = subprojects/wayland-logout
 	url = https://github.com/soreau/wayland-logout
+[submodule "subprojects/gtk4-layer-shell"]
```
ammen99, line 7: Shouldn't we remove the gtk-layer-shell submodule (which is for gtk3)?

src/panel/panel.cpp
```diff
@@ -108,47 +76,44 @@ class WayfirePanel::impl
...
-            sigc::mem_fun(this, &WayfirePanel::impl::on_delete));
-    }
-
-    bool on_delete(GdkEventAny *ev)
-    {
-        /* We ignore close events, because the panel's lifetime is bound to
-         * the lifetime of the output */
-        return true;
+        // window->signal_close_request().connect(
+        // sigc::mem_fun(*this, &WayfirePanel::impl::on_delete));
```
ammen99: if this is not needed anymore we cna just delete the code, no need to leave it commented :)

src/panel/panel.cpp
```diff
@@ -315,6 +291,24 @@ class WayfirePanel::impl
   public:
     impl(WayfireOutput *output) : output(output)
     {
+        // Intentionally leaking feels bad.
```
ammen99: I wonder whether these shouldn't be static variables. We don't need one instance per output, right? This creates new objects every time we plug an output, but if they were static variables, they would be initialized just once for all outputs.

src/panel/panel.cpp
```diff
@@ -315,6 +291,24 @@ class WayfirePanel::impl
...
+        new CssFromConfigString("panel/background_color", ".wf-panel{background-color:", ";}");
+        new CssFromConfigBool("panel/battery_icon_invert", ".battery image{filter:invert(100%);}", "");
+        new CssFromConfigBool("panel/network_icon_invert_color", ".network-icon{filter:invert(100%);}", "");
+
+        // People will probably need to update sizes to have a measure
+        // 16px, 1.1rem, 1em .
+        // So on
+        new CssFromConfigFont("panel/battery_font", ".battery {", "}");
+        new CssFromConfigFont("panel/clock_font", ".clock {", "}");
+
```
ammen99: +1000 for the idea to translate .ini to css, I really like the solution :)

src/panel/widgets/window-list/window-list.cpp
```diff
@@ -40,59 +40,6 @@ void WayfireWindowListBox::set_top_x(int x)
...
-    std::vector<Gtk::Widget*> result;
-    for (auto& child : children)
-    {
-        result.push_back(Glib::wrap(child));
-    }
-
-    return result;
-}
-
-void WayfireWindowListBox::forall_vfunc(gboolean value, GtkCallback callback, gpointer callback_data)
```
ammen99, line 68: The purpose of this code was to ensure that we can drag and drop toplevels in the list to rearrange them. I assume that custom drawing is no longer as simple as this code implements it?

If so, could we make use of gtk drag and drop to rearrange toplevels, potentially showing a drag icon? Currently, drag and drop does rearrange the toplevels but there is no animation and no indication of what is going on, which I personally find a bit confusing.

In addition, at the end of the drag and drop, the dragged view gets a special status (probably 'active drag and drop') and remains like that, so we have to unset somethinig probably.

ammen99 (changes requested):
I looked over the changes, really good work :) My focus was the structure, haven't looked at every single UI call in detail but I assume we will find bugs like that during testing. A few things that I noticed or questions that I have.

.gitmodules
trigg: Unless I've missed something, that is GTK4, not 3

src/panel/panel.cpp
trigg: Yeah, that's a running bad habit of mine. Sorry!

.gitmodules
ammen99: Sure, this is the good one, but the 3 version is also there in the gitmodules (unless I am mistaken)

trigg: So it is, but `.git/config` had the right of it, and it seemed to be ignoring `.gitmodules`. Strange, but okay.

src/panel/panel.cpp
trigg: I've moved them up into the App class, during on_activate. This is a much better position for them

src/panel/panel.cpp
trigg: It seemed like the thing to do at the time.

I've had to re-write the font handling as it was broken again, and it hadn't been handling people using fonts like `NotoSans 10` to assume `10pt`. It now assumes `pt` if no unit is given from `px` `rem` or `em`

src/panel/widgets/window-list/window-list.cpp
trigg: Theoretically, we could subclass a LayoutManager, and write our own logic on size and location of children.

In reality I can't make it compile when I try.

trigg: This is done now.

trigg:
> I would simply squash the PR ...

I absolutely would squash this. I could think of no sensible and meaningful way to split this.

> `/<prefix>/share/..` or so, and load them with lower priority ...

Working on a commit for this. It already was a lower priority, but loading in-situ fixes #254 too

ammen99 (approved):
LGTM, thanks a lot, really appreciate your work and responsiveness!

PR #301, opened 2025-07-19
Language widget for panel

WShad:
Adds language widget displaying current keyboard layout language.

Corresponding proto implementation PR: https://github.com/WayfireWM/wayfire/pull/2742

### DIFF
--- a/src/panel/widgets/language.cpp
+++ b/src/panel/widgets/language.cpp
@@ -1,0 +1,112 @@
+#include <cstdint>
+#include <glibmm.h>
+#include <iostream>
+#include <map>
+#include <string>
+#include <vector>
+#include <xkbcommon/xkbregistry.h>
+#include "language.hpp"
+#include "gtkmm/button.h"
+#include "libutil.a.p/wayfire-shell-unstable-v2-client-protocol.h"
+#include "sigc++/functors/mem_fun.h"
+
+void WayfireLanguage::init(Gtk::HBox *container)
+{
+    // button = std::make_unique<Gtk::Button>("panel");
+
+    button.get_style_context()->add_class("language");
+    button.add(label);
+    button.get_style_context()->add_class("flat");
+    button.get_style_context()->remove_class("activated");
+    button.signal_clicked().connect_notify(sigc::mem_fun(this, &WayfireLanguage::next_language));
+    button.show();
+    label.show();
+
+    update_label();
+
+    container->pack_start(button, false, false);
+}
+
+bool WayfireLanguage::update_label()
+{
+    if (current_language >= available.size()) {
+        return false;
+    }
+
+    label.set_text(available[current_language].ID);
+    return true;
+}
+
+static void keyboard_lang_manager_current_layout(void *data, struct zwf_keyboard_lang_manager_v2 *keyboard_lang_manager, const uint32_t id)
+{
+    auto wf_language = (WayfireLanguage*)data;
+    wf_language->set_current(id);
+    wf_language->update_label();
+}
+
+static void keyboard_lang_manager_available_layouts(void *data, struct zwf_keyboard_lang_manager_v2 *keyboard_lang_manager, wl_array *layouts)
+{
+    std::vector<Language> languages;
+    std::map<std::string, uint32_t> names;
+    char *elem = (char *) layouts->data;
+    uint32_t index = 0;
+    while (elem < (char *) layouts->data + layouts->size) {
+        names[std::string(elem)] = index;
+        languages.push_back(Language{
+            .Name = elem,
+            .ID = "",
+        });
+        size_t length = strlen(elem) + 1;
+        elem += length;
+        index++;
+    }
+
+    auto context = rxkb_context_new(RXKB_CONTEXT_NO_FLAGS);
+    rxkb_context_parse_default_ruleset(context);
+    auto rlayout = rxkb_layout_first(context);
+    for (; rlayout != NULL; rlayout = rxkb_layout_next(rlayout)) {
+        auto descr = rxkb_layout_get_description(rlayout);
+        auto name = names.find(descr);
+        if (name != names.end()) {
+            languages[name->second].ID = rxkb_layout_get_brief(rlayout);
+        }
+    }
+
+    auto wf_language = (WayfireLanguage*)data;
+    wf_language->set_available(languages);
+}
+
+static zwf_keyboard_lang_manager_v2_listener listener = {
+    .current_layout = keyboard_lang_manager_current_layout,
+    .available_layouts = keyboard_lang_manager_available_layouts,
+};
+
+
+void WayfireLanguage::set_current(uint32_t index)
+{
+    current_language = index;
+}
+
+void WayfireLanguage::set_available(std::vector<Language> languages)
+{
+    available = languages;
+}
+
+void WayfireLanguage::next_language()
+{
+    uint32_t next = current_language + 1;
+    if (next >= available.size())
+    {
+        next = 0;
+    }
+
+    zwf_keyboard_lang_manager_v2_set_layout(keyboard_lang_manager, next);
+}
+
+WayfireLanguage::WayfireLanguage(zwf_keyboard_lang_manager_v2 *keyboard_lang_manager): keyboard_lang_manager(keyboard_lang_manager)
+{
+    zwf_keyboard_lang_manager_v2_add_listener(keyboard_lang_manager, &listener, this);
+}
+
+WayfireLanguage::~WayfireLanguage()
+{}
--- a/src/util/wf-shell-app.hpp
+++ b/src/util/wf-shell-app.hpp
@@ -1,6 +1,7 @@
 #ifndef WF_SHELL_APP_HPP
 #define WF_SHELL_APP_HPP
 
+#include <memory>
 #include <set>
 #include <string>
 #include <wayfire/config/config-manager.hpp>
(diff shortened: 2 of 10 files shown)


WShad:
@ammen99, better?))

src/util/wf-shell-app.hpp
```diff
@@ -66,6 +67,7 @@ class WayfireShellApp
     int inotify_css_fd;
     wf::config::config_manager_t config;
     zwf_shell_manager_v2 *wf_shell_manager = nullptr;
+    WayfireIPC *ipc = nullptr;
```
NamorNiradnug: I would suggest making `WayfireIPC` a singleton so that it won't be initialized until needed.

I usually do it like this:
- `WayfireIPC` doesn't have a public ctor
- has private `static std::weak_ptr<WayfireIPC> instance_`
- has public `static std::shared_ptr<WayfireIPC> Instance()` which simply returns `instance_.lock()`
- has public `static std::shared_ptr<WayfireIPC> Open()` (or something, maybe you will come up with a better name) which initialize the `instance_` if it's expired

`Instance()` should be called in `reload_widgets` so that we don't close-and-open the socket on every reload.

Probably `Lock()` would be a better name for `Instance()`, but I would still call it `Instance` for consistency with existing singletons.

See https://github.com/WayfireWM/wf-shell/blob/master/src/panel/widgets/tray/watcher.hpp#L18-L31 and https://github.com/WayfireWM/wf-shell/blob/master/src/panel/panel.cpp#L268.

src/util/wf-ipc.hpp
```diff
@@ -0,0 +1,45 @@
...
+class IIPCSubscriber {
+    public:
+        virtual void on_event(nlohmann::json) = 0;
+};
+
+using response_handler = std::function<void(nlohmann::json)>;
+
+class WayfireIPC
+{
+        int socket_fd;
```
NamorNiradnug: Have you considered using `Gio`'s `SocketClient` instead of handling it manually? I believe it would simplify the code.

ammen99:
@WShad sorry for the delay, we were finishing the gtk4 port .. if you still have interest, please update this PR to gtk4 as well.

src/panel/widgets/language.cpp
```diff
@@ -0,0 +1,127 @@
...
+void WayfireLanguage::init(Gtk::Box *container)
+{
+    button.get_style_context()->add_class("language");
+    button.get_style_context()->add_class("flat");
+    button.get_style_context()->remove_class("activated");
+    button.signal_clicked().connect(sigc::mem_fun(*this, &WayfireLanguage::next_layout));
+    button.show();
+
+    ipc->subscribe(this, {"keyboard-modifier-state-changed"});
+    ipc->send("{\"method\":\"wayfire/get-keyboard-state\"}", [=] (json_t data)
```
ammen99: I find your design very interesting with this send and the callback for the response. One thing I am not very sure about however is this, what deregisters the callback if the WayfireLanguage class is destroyed while waiting for the reply?

src/util/wf-ipc.cpp
```diff
@@ -0,0 +1,297 @@
...
+        connection->get_socket()->get_fd(),
+        Glib::IO_OUT);
+}
+
+void WayfireIPC::write_stream(const std::string& message)
+{
+    try {
+        writing = true;
+        uint32_t length = message.size();
+        auto all_data   = std::make_shared<std::string>((char*)&length, 4);
```
ammen99: why do we need a shared ptr here? it seems to me we just pass it to write_all_async, and the `all_data` capture isn't really used.

WShad: Because i need to pass in pointer to write_all_async, and that pointer must be valid until data completely wrote, but capture can't protect pointer from being dangling, so i need something to hold it in place.

From documentation:

> Note that no copy of @a buffer will be made, so it must stay valid until @a slot is called.

src/panel/widgets/language.cpp
WShad: > I find your design very interesting with this send and the callback for the response

Thanks! I'm glad to hear that)

> what deregisters the callback if the WayfireLanguage class is destroyed while waiting for the reply?

Nice point. Didn't think about it.

src/util/wf-ipc.cpp
ammen99: Alright, but then please add this as a comment, so that future readers can also understand what is going on.

src/panel/widgets/language.cpp
ammen99: One idea would be to somehow connect the full `ipc->send()` version with the IPC subscriber interface, then have the IPC subscriber interface deregister itself from the IPC socket automatically. This is what we do for signals in Wayfire itself. Maybe the implementation will give you some inspirations, but feel free to come up with a different idea if you have a better one :)

https://github.com/WayfireWM/wayfire/blob/master/src/api/wayfire/signal-provider.hpp
https://github.com/WayfireWM/wayfire/blob/8a3e1cae5beda079060d7c5725812bdc589a6ae3/src/core/object.cpp#L7-L59

WShad: Resolved. Now widgets must have their own IPC clients, they deregister themselves and all replies to that clients will be safely ignored.

src/util/wf-ipc.cpp
WShad: Done 👍

src/util/wf-ipc.cpp
```diff
@@ -0,0 +1,346 @@
...
+    sig_connection = Glib::signal_io().connect(
+        sigc::mem_fun(*this, &WayfireIPC::receive),
+        connection->get_socket()->get_fd(),
+        Glib::IO_IN);
+}
+
+WayfireIPC::~WayfireIPC()
+{
+    disconnect();
+    LOGI("IPC destroyed");
```
ammen99: i don't think we need this message in 'production' :)

src/util/wf-ipc.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+bool WayfireIPC::receive(Glib::IOCondition cond)
+{
+    try {
+        ssize_t received = 0;
+
+        while (connection->get_socket()->get_available_bytes() > 0)
+        {
+            received = input->read(&length, 4);
```
ammen99: I prefer to have length as a local variable, it is not some 'state' which needs to be preserved across method calls.

src/util/wf-ipc.cpp
```diff
@@ -0,0 +1,346 @@
...
+            if (received == -1)
+            {
+                LOGE("Receive message length failed");
```
ammen99: when we log errors, it would make sense to also indicate where they are coming from, for example: "Error in Wayfire IPC communication: failed to receive message length"

ammen99: (also applies to all the other LOGE calls in this file / PR

src/util/wf-ipc.cpp
ammen99: Also instead of `4`, consider using `sizeof(length)`, that way I think it is way clearer why we read those bytes separately :)

ammen99: Another thing: when calling `read()`, I we are not guaranteed to get the full number of bytes we request. In particular, if we have a longer ipc message, I think read **could** stop in the middle, return just some part of the message. At least the underlying read() syscall works this way. Example of what Wayfire does to combat this: https://github.com/WayfireWM/wayfire/blob/9c427024e849b0c676184c1af3c82b38f1f8fb67/plugins/ipc/ipc.cpp#L171-L285

ammen99 (changes requested):
getting closer to perfection :) a few smaller items that i noticed, otherwise lgtm

src/util/wf-ipc.cpp
WShad: Can't actually confirm that. I experimented a lot and this is a results:

- under the pressure of data socket can get buffer overflow, if not read frequent enough
- if n bites written - n bites can be read (not like in a tcp connection)
- if server lags on read or write - client disconnected immediately 

Can't reproduce situation where was read less than was written.
Using this IPC for three months, fixed some bugs, have no problems.

src/util/wf-ipc.cpp
WShad: Disagree, but - ok

src/util/wf-ipc.cpp
WShad: > I prefer to have length as a local variable, it is not some 'state' which needs to be preserved across method calls.

Will fix. Legacy of prev impl)

ammen99: > * under the pressure of data socket can get buffer overflow, if not read frequent enough

That's exactly the issue! It can be that the buffer can only accommodate the message partially.

I mean in practice I expect this to not cause any issues, so maybe we could simply leave this as a TODO comment in the code and fix it when somebody actually complains.

ammen99: In any case would be good to at least check what is going on, so instead of errorring on `n_bytes_read == 0`, we need to error on `n_bytes_read != n_bytes_wanted`

ammen99:
@WShad For the CI to work, I think you need to add cmake to the list of dependencies, or to install yyjson as a dependency so that we don't have to build it as a submodule. Also you'll need to fix the codestyle. Otherwise LGTM, good work!

WShad:
Clean now: https://github.com/WShad/wf-shell/actions/runs/20010484337

WShad:
I don't understand... How? My action completed successfully, they are the same - how it even can be?

birdie-github:
Does it support a language emoji? If not, would be very nice to have along with an option to either show:
 * text only (EN)
 * text and language emoji (EN 🇬🇧)
 * just language emoji (🇬🇧)

Issue #259, opened 2024-05-22
moveable panel


brikler:
salve,

i am using wayfire on a notebook, and the screen is 16:9, so the panel on top, or bottom, makes the usable space smaller then necessary, but but on the right, or left side are, more or less, unused space… in short, i prefer the panel on the right side of the screen…maybe you can make this happen :)

sincerely :)

anoraktrend:
this is possible!

AKArien0:
Hi,
just wanted to chime in to say i’m doing that ! and adding a few other options while i’m at it.

AKArien:
Hi, with #306 merged, it now exists ! Can be closed

PR #306, opened 2025-11-01
Added options for panel/dock positions and finer autohide configuration

AKArien:
- Adjusted the naming of a few options in adequation with new features
- Added options to control the delay before doing the showing/hiding animations
- Added options to place the panel and dock on the left and right edges of the screen
- Added minimal width option
- Added option to force spanning the full edge
- Option to force the panel to be centered by the middle box instead of the total width. This will lead to more spaced out elements in the side that takes less space

For the panel, most widgets have a layout adaptation for the vertical panel. Exceptions are network and window-list, because i couldn’t figure out a clean or good way to do it are. Textual widgets (clock, command-output and arguably both of the previously mentionned unchanged) don’t have special adaptations, the formatting for the text they output should be adapted to account for it. Maybe it would be a good idea to add examples to the default config file/wiki.

Breaking stuff :
- dock/dock_height is replaced by dock/minimal_height

Adresses #259

### DIFF
--- a/metadata/dock.xml
+++ b/metadata/dock.xml
@@ -12,8 +12,19 @@
 		<default>true</default>
 	</option>
 	<option name="autohide_duration" type="int">
-		<_short>Autohide duration</_short>
+		<_short>Autohide animation duration</_short>
+		<_long>Time (in milliseconds) the dock takes to show and hide</_long>
 		<default>300</default>
+	</option>
+	<option name="autohide_show_delay" type="int">
+		<_short>Autohide show delay</_short>
+		<_long>Amount of time (in milliseconds) the cursor needs to be in the zone before showing.</_long>
+		<default>300</default>
+	</option>
+	<option name="autohide_hide_delay" type="int">
+		<_short>Autohide hide delay</_short>
+		<_long>Amount of time (in milliseconds) the cursor needs to be out of the zone before hiding.</_long>
+		<default>500</default>
 	</option>
 	<option name="position" type="string">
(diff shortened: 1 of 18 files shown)


AKArien:
@trigg hey there, taking the liberty to ping you to ask how much breaking is ok.
notably, this is currently breaking dock_height to unify minimal sizes between panel and dock. With it introducing a vertical panel, i currently changed the description of widgets_left and right to indicate wether it becomes the top or bottom. Soreau also suggested to rename them as start/end, which i do see the value in, although it’s a bit unclear at a first reading which is which.
And hey, if we are breaking things… i am once again asking to change launchers_size into launchers_icon_size :)
He also suggested splitting this pr in a few smaller ones, which i will let you be the arbiter of soundness of.
Finally, if you have any idea on how to make the reflowing dock look good on the bottom, i’d like to hear :)
And, ofc, comment on the code problems ! Notably reloading the layout of widgets for a vertical panel currently has a lot of copypaste, but i’m not sure what is the best way to cut down on it.

AKArien:
Just a note that i’d like to make some changes accounting for things that have been merged recently (notably, 328, 339, and 343 if it gets merged)

For the deprecation, we are waiting for a wf-config mechanism, right ? Has there been discussion/decisions on how it should be done ?

trigg (approved):
A terse read! Very far reaching.

That said I believe it is worth the large change to get these features.

Lined up for merge, breaking config be damned. We need to ensure the message goes into next release.

PR #307, opened 2025-11-02
Added wireplumber mixer widget

AKArien:
The widget gives an interface for controlling the volume and muting of sinks, sources and output streams, as well as selection of the default sinks and sources (input/output devices).
By default, upon changes to a volume / mute, it pops up an interactible bubble where volume/mute can be changed again.
Has scroll to change volume and possibility of showing the controls of only a single track at a time, through the « face/quick action target »
Has the following settings :
- Popup on change : enable or disable the « popup on change » behavior. If on, the bubble described above will appear when the « face » volume or mute changes.
- Timeout for hiding popups
- Length of sliders
- Scroll sensitivity (defaults to 1), and inversion
- The buttons to mute and select default are on the right by default, but can be switched to the left
- Selection of what should be the « quick target » for muting and to be shown for a quick control with the corresponding actions : last changed volume, default sink, default source.
- Assignment of view of the full mixer, quick target and muting between left, middle and right click (mute not available on left click)
- icon size

It’s build is behind a new build option and uses the wireplumber C library.
Code for icon selection and the custom gtk scale have been moved out of the pulseaudio volume widget for re-use.
This PR slightly alters the volume widget, by changing the « volume out of range » icon the « emblem-unreadable » icon instead of « audio-volume-muted »
Additionally, the compile-time warning that was emitted if libpulse was not found and the volume widget not built has been moved to a meson notice.

### DIFF
--- a/wf-shell.ini.example
+++ b/wf-shell.ini.example
@@ -169,6 +169,9 @@ wp_invert_scroll = false
 # Set to true to show icons (mute, set default) on the left
 wp_icons_on_left = false
 
+# Spacing between the controls of the mixer
+wp_spacing = 8
+
 # Actions to be tied to left, right and middle click on the widget
 # Can be:
 #   - show_mixer: show the whole mixer


src/panel/panel.cpp
```diff
@@ -164,6 +167,17 @@ class WayfirePanel::impl
 #endif
         }
 
+        if (name == "wireplumber")
+        {
+#ifdef HAVE_WIREPLUMBER
+            return Widget(new WayfireWireplumber());
+#else
+    #warning "Wireplumber not found, mixer widget will not be available."
```
ammen99: I think this warning, if emitted at all, should be emitted from meson.build, not from the source file.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,1006 @@
...
+        control->handle_config_reload();
+    }
+}
+
+void WayfireWireplumber::init(Gtk::Box *container)
+{
+    // sets up the « widget part »
+
+    button = std::make_unique<WayfireMenuButton>("panel");
+    button->get_style_context()->add_class("volume");
```
ammen99: I think we should use a different class for wireplumber.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,1006 @@
...
+    sources_box.set_orientation(r2);
+    sources_box.append(*new Gtk::Label("Input devices"));
+    sources_box.append(*new Gtk::Separator(r2));
+    streams_box.set_orientation(r2);
+    streams_box.append(*new Gtk::Label("Audio streams"));
+    streams_box.append(*new Gtk::Separator(r2));
+
+    /* Setup popover */
+    popover->set_child(master_box);
+    popover->get_style_context()->add_class("volume-popover");
```
ammen99: same here, should be a different name for proper css styling support.

src/panel/widgets/wireplumber.hpp
```diff
@@ -0,0 +1,147 @@
...
+     */
+    void check_set_popover_timeout();
+
+    void reload_config();
+    void handle_config_reload();
+
+    virtual ~WayfireWireplumber();
+};
+
+namespace WpCommon
```
ammen99: I wonder whether this really has to be in the .hpp file, it is just an implementation detail and used only in wireplumber.cpp, right?

AKArien:
Should be good !'m not running intitulé any bugs in my own use.
The one thing i’m not quite happy about is the hacky way of « keeping the popover out but swapping it’s content » on click by closing it right before, so it gets opened again by the click event (line 469/490).
It might also help with readability to split it into multiple files ?

src/panel/widgets/wireplumber.hpp
```diff
@@ -0,0 +1,152 @@
...
+{
+static WpCore *core = nullptr;
+static WpObjectManager *object_manager;
+static WpPlugin *mixer_api;
+static WpPlugin *default_nodes_api;
+
+static std::vector<WayfireWireplumber*> widgets;
+
+static void init_wp();
```
NamorNiradnug: Do not define `static` functions and objects in headers. These functions should go into `wireplumber.cpp`.

Also I don't like the global variable above...

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
+#include <fontconfig/fontconfig.h>
+#include <gtkmm.h>
+#include <iostream>
+#include <glibmm.h>
+#include "gio/gio.h"
```
NamorNiradnug: Please use `<...>` for library headers.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+    button.set_child(volume_icon);
+
+    attach(label, 0, 0, 2, 1);
+    attach(button, 1, 1, 1, 1);
+    attach(scale, 0, 1, 1, 1);
+
+    mute_conn = button.signal_toggled().connect(
+        [this, id] ()
+    {
+        GVariantBuilder gvb = G_VARIANT_BUILDER_INIT(G_VARIANT_TYPE_VARDICT);
```
NamorNiradnug: Please use `Glib::Variant` instead of C glib functions.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+{
+    VolumeLevel current = volume_icon_for(get_scale_target_value());
+
+    if (is_muted())
+    {
+        volume_icon.set_from_icon_name("audio-volume-muted");
+        return;
+    }
+
+    std::map<VolumeLevel, std::string> icon_name_from_state = {
```
NamorNiradnug: `icon_name_from_state` can be made `static const`.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+}
+
+void WfWpControl::handle_config_reload()
+{
+    update_gestures();
+}
+
+WfWpControl*WfWpControl::copy()
+{
+    WfWpControl *copy = new WfWpControl(object, parent);
```
NamorNiradnug: Do not return objects by raw pointers. Either use `std::unique_ptr` or return by value.

~~Also why don't you just use the default copy constructor?~~ Ah, because there are callbacks that store a pointer to the object. Then you probably want to make the object non-movable.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+    reload_config();
+
+    // boxes hierarchy and labeling
+    auto r1 = Gtk::Orientation::HORIZONTAL;
+    auto r2 = Gtk::Orientation::VERTICAL;
+
+    master_box.set_orientation(r1);
+    // TODO : only show the boxes which have stuff in them
+    master_box.append(sinks_box);
+    master_box.append(*new Gtk::Separator(r1));
```
NamorNiradnug: Memory leak here. Just `append(Gtk::Separator(r1))`.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+{
+    VolumeLevel current = volume_icon_for(face->get_scale_target_value());
+
+    if (!face || face->is_muted())
+    {
+        main_image.set_from_icon_name("audio-volume-muted");
+        return;
+    }
+
+    std::map<VolumeLevel, std::string> icon_name_from_state = {
```
NamorNiradnug: The same map is defined in another place. Don't duplicate code.

Probably `icon_name_from_state` function should be OK.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+    {
+        auto it = widget->objects_to_controls.find((WpPipewireObject*)object);
+        if (it == widget->objects_to_controls.end())
+        {
+            // shouldn’t happen, but checking to avoid weird situations
+            return;
+        }
+
+        WfWpControl *control = it->second;
+        Gtk::Box *box = (Gtk::Box*)control->get_parent();
```
NamorNiradnug: `auto *`

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+#include "wp/proxy-interfaces.h"
+#include "wp/proxy.h"
+
+#include <memory>
+#include <pipewire/keys.h>
+
+#include "wireplumber.hpp"
+
+
+enum VolumeLevel
```
NamorNiradnug: The same `enum` is already defined in another place. Reuse it.

src/util/meson.build
```diff
@@ -6,6 +6,7 @@ util = static_library(
         'wf-autohide-window.cpp',
         'wf-popover.cpp',
         'css-config.cpp',
+		'animated-scale.cpp'
```
NamorNiradnug: Use spaces here

meson.build
```diff
@@ -33,6 +34,10 @@ if libpulse.found()
 	add_project_arguments('-DHAVE_PULSE=1', language: 'cpp')
 endif
 
+if wireplumber.found()
+	add_project_arguments('-DHAVE_WIREPLUMBER=1', language: 'cpp')
```
NamorNiradnug, line 40: Use spaces here

src/panel/widgets/wireplumber.hpp
```diff
@@ -0,0 +1,152 @@
...
+    sigc::connection volume_changed_signal;
+
+    bool gestures_initialised = false;
+    std::shared_ptr<Gtk::GestureClick> left_click_gesture, middle_click_gesture, right_click_gesture;
+    sigc::connection left_conn, middle_conn, right_conn;
+
+  public:
+    void init(Gtk::Box *container) override;
+
+    WfOption<double> scroll_sensitivity{"panel/wp_scroll_sensitivity"};
```
NamorNiradnug: The fields should be private.

src/panel/widgets/wireplumber.hpp
```diff
@@ -0,0 +1,152 @@
...
+    WfOption<bool> invert_scroll{"panel/wp_invert_scroll"};
+    WfOption<bool> popup_on_change{"panel/wp_popup_on_change"};
+
+    FaceChoice face_choice;
+
+    std::unique_ptr<WayfireMenuButton> button;
+    Gtk::Popover *popover;
+
+    /*
+     *   the « face » is the representation of the audio channel that shows it’s volume level on the widget
```
NamorNiradnug: Please don't use unicode characters in code and comments. `« face »`→ `"face"`

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+                return def_id == bound_id;
+            });
+            if (is_default)
+            {
+                ctrl->set_def_status_no_callbk(true);
+                continue;
+            }
+
+            // if the control is not for a sink or source (non WfWpControlDevice), don’t try to set status
+            const gchar *type = wp_pipewire_object_get_property(obj, PW_KEY_MEDIA_CLASS);
```
NamorNiradnug: Make this variable of type `std::string_view`. Then you can compare it with strings  using `==` without all these `g_strcmp0`s.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+    const gchar *type = wp_pipewire_object_get_property(obj, PW_KEY_MEDIA_CLASS);
+
+    for (auto widget : widgets)
+    {
+        WfWpControl *control;
+        Gtk::Box *which_box;
+        if (g_strcmp0(type, "Audio/Sink") == 0)
+        {
+            which_box = &(widget->sinks_box);
+            control   = new WfWpControlDevice(obj, widget);
```
NamorNiradnug: Don't manage memory with `new/delete`.

Probably `objects_to_control` should own the objects through `std::unique_ptr`.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+        NULL);
+
+    wp_core_install_object_manager(core, object_manager);
+}
+
+void WpCommon::on_object_added(WpObjectManager *manager, gpointer object, gpointer data)
+{
+    // adds a new widget to the appropriate section
+
+    WpPipewireObject *obj = (WpPipewireObject*)object;
```
NamorNiradnug: `auto *`

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+
+    const gchar *type = wp_pipewire_object_get_property(obj, PW_KEY_MEDIA_CLASS);
```
NamorNiradnug: `std::string_view`

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+}
+
+void WayfireWireplumber::reload_config()
+{
+    WfOption<std::string> str_face_choice{"panel/wp_face_choice"};
+    WfOption<std::string> str_wp_left_click_action{"panel/wp_left_click_action"};
+    WfOption<std::string> str_wp_right_click_action{"panel/wp_right_click_action"};
+    WfOption<std::string> str_wp_middle_click_action{"panel/wp_middle_click_action"};
+
+    if ((std::string)str_face_choice == (std::string)"last_change")
```
NamorNiradnug: Don't cast a string constant to `std::string` for comparison. `std::string` has `operator==` with `const char *`.

Also please use `.value()` method of `WfOption` instead casting it to `std::string`: it looks better and would allow to optimize unnecessary copy if `value()` would return a reference instead of copy (it currently doesn't, but I think we should fix it).

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,984 @@
...
+        return true;
+    }, true);
+    scroll_gesture->set_flags(Gtk::EventControllerScroll::Flags::VERTICAL);
+    button->add_controller(scroll_gesture);
+
+    // config options
+    reload_config();
+
+    // boxes hierarchy and labeling
+    auto r1 = Gtk::Orientation::HORIZONTAL;
```
NamorNiradnug: Please don't use such aliases. It decreases code readability.

ammen99:
> It might also help with readability to split it into multiple files ?

Feel free to do so if you find it better. We already do that for window-list, simply create a subdirectory and place all files in it.

meson.build
AKArien: Now, apologies if i am somehow getting this wrong, but it seems other indentation in this file uses tabs ? such as the above libpulse check

src/panel/widgets/wireplumber.hpp
AKArien: > Also I don't like the global variable above...

As far as i see, the alternative is connect the signals with the WayfireWireplumber as user data, for each widget ? Is that better or worse ?

src/panel/widgets/wireplumber.cpp
AKArien: So, I’m sorry but for some reason i can’t seem to figure it out… Do you know how to do it ?
My best shot is something like this :
```c++
        auto dict = Glib::VariantDict::create();
        dict->insert_value_variant("mute", Glib::Variant<bool>::create(button.get_active()));
        gboolean res FALSE;
        g_signal_emit_by_name(WpCommon::mixer_api, "set-volume", id, dict->gobj(), &res);
```
But it aborts on a type info check assertion and i seem to phenomenally fail at looking anything up.

src/panel/widgets/wireplumber.hpp
AKArien: Actually, WfWpControl accesses scroll sensitivity through it’s parent right now (lines 136/140)

metadata/panel.xml
```diff
@@ -249,6 +249,99 @@
...
+		</_long>
+		<default>true</default>
+	</option>
+	<option name="wp_display_timeout" type="double">
+		<_short>Display timeout</_short>
+		<default>2.5</default>
+	</option>
+	<option name="wp_scroll_sensitivity" type="double">
+		<_short>Scroll sensitivity</_short>
+		<default>1</default>
```
ammen99, line 287: We could set a <min> value for this (0?) Otherwise, we could allow negative values but then we can remove the invert option.

src/panel/widgets/wireplumber.cpp
ammen99: I am no dbus expert but take a look at this other PR, maybe it would give you an inspiration: https://github.com/WayfireWM/wf-shell/pull/311#issuecomment-3522202783

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,983 @@
...
+#include <wp/proxy-interfaces.h>
+#include <wp/proxy.h>
+
+#include <pipewire/keys.h>
+#include <string_view>
+
+#include "volume-level.hpp"
+#include "wireplumber.hpp"
+
+namespace WpCommon
```
ammen99: This one seems like it should be a singleton class.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,983 @@
...
+    /* Setup layout */
+    container->append(*button);
+    button->set_child(main_image);
+
+    /*
+     * If the core is already set, we are another widget, wether on another monitor or on the same wf-panel.
+     * We re-use the core, manager and all other objects
+     */
+
+    WpCommon::widgets.push_back(this);
```
ammen99: Once you have a singleton class for the WpCommon stuff, you can move all this code in a method there (`add_widget` / `remove_widget`)

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,983 @@
...
+    DEFAULT_SINK,
+    DEFAULT_SOURCE,
+};
+
+static const gchar *DEFAULT_NODE_MEDIA_CLASSES[] = {
+    "Audio/Sink",
+    "Audio/Source",
+};
+
+static VolumeLevel volume_icon_for(double volume)
```
ammen99: This is the same code as in `volume.cpp`, I think we can avoid duplication here. `pa_volume_t` is a uint32_t, so we can simply use `double` for the type everywhere.

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,983 @@
...
+        right_conn = right_click_mute->signal_pressed().connect(mute_action);
+    }
+}
+
+void WfWpControl::handle_config_reload()
+{
+    update_gestures();
+}
+
+std::unique_ptr<WfWpControl> WfWpControl::copy() // for the face handling
```
ammen99: I don't pretend to have the entirety of the code in my head, but could we use `std::shared_ptr` instead of `unique_ptr`s here and not have to copy the current face? Do we really need a copy or just a pointer to it?

src/panel/widgets/wireplumber.cpp
```diff
@@ -0,0 +1,983 @@
...
+    popover_timeout.disconnect();
+
+    popover_timeout = Glib::signal_timeout().connect(sigc::bind(sigc::mem_fun(*this,
+        &WayfireWireplumber::on_popover_timeout), 0), timeout * 1000);
+}
+
+void WayfireWireplumber::reload_config()
+{
+    // big matching operation
+    WfOption<std::string> str_face_choice{"panel/wp_face_choice"};
```
ammen99: These can be marked as static to avoid reloading them every time. Their values will change automatically when the config changes.

src/panel/widgets/wireplumber.cpp
AKArien: Oh hey, i just changed that :P

src/panel/widgets/wireplumber.cpp
AKArien: As far as i understand, if we were to not copy, we would then be moving the control (gtk) widget to the face, and then we would be moving it back to the appropriate box in the popover. The problem with this is that it would mess up the order everytime, which i personally found to be annoying to use. This could be remedied having the order of the controls be more « properly » enforced, and we could also have other ways of sorting.
It would also lead to vacant face when displaying the full mixer, since we’d be moving it back to the mixer. Not strictly a bad thing, but i’m not a fan.

metadata/panel.xml
AKArien: I’m sorry, i’m not quite sure what you mean ?
As for invert, we could indeed remove it, but i included it because i can’t help but feel it is way nicer as a user to have a toggle rather than making the value negative, at least when using wcm. Should it go ?

src/panel/panel.cpp
AKArien: Sure, based myself on the pulseaudio volume widget. Should i change it as well ?

src/panel/widgets/wireplumber.cpp
AKArien: i’ve actually been writing a display brightness control widget that ends up having roughly the same thing again. Should i preemptively make it even more generic to account for it (or other stuff that might have be similar), and if it’s done soon enough, should i add it to this pr or make another one ?

src/panel/panel.cpp
ammen99: Sounds good!

src/panel/widgets/wireplumber.cpp
ammen99: Sorry, this was a leftover from a previous review attempt.

metadata/panel.xml
ammen99: > I’m sorry, i’m not quite sure what you mean ?

You can set minimum and maximum values for integer options. `<min>0</min>` for example. I am fine with either retaining both options and setting the minimum to 0, or merging them together (in which case minimum makes no sense). Pick whichever one you like more.

src/panel/widgets/wireplumber.cpp
ammen99: Better leave it for later, lets finish this PR first and then worry about the next one.

src/panel/widgets/wireplumber.cpp
ammen99: I see.

src/panel/panel.cpp
AKArien: > if emitted at all

Changed it to be just a message, assuming that’s what you’d want.

AKArien:
@ammen99 i think i have addressed everything pointed out here, but i have a few little ui bugs to fix.

There’s also an issue that i really don’t understand and would like you to keep an eye on in your testing / review : sometimes, under circumstances i cannot nail down, the tray or the menu widgets will segfault, with gdb showing a backtrace in the Watcher destructor for the tray and the toggle_menu method for menu. This never happens without wireplumber in my config, and removing it is always an immediate fix. My widgets are usually like this :
```
widgets_center = separator1 menu separator1
widgets_left = tray wireplumber
widgets_right = notifications launchers battery clock
```
I hardly see how this could be related to my widget, but it seems i’m missing something ? Since removing wireplumber always stops it from happening. Again, i’m sorry if this has very little detail, but it is litterally all that i could find to be consistent.

ammen99:
@AKArien I would strongly recommend compiling wf-shell with address sanitizer in this case. It is likely that you see the result of a memory corruption somewhere and address sanitizer very often catches issues like these.

metadata/panel.xml
```diff
@@ -249,6 +249,106 @@
 	</option>
 	</group>
 	<group>
+	<_short>Wireplumber mixer</_short>
+	<option name="wp_popup_on_change" type="bool">
+		<_short>Pop up on change</_short>
+		<_long>
+			Wether to show a popup with the new values upon a change to the quick action target
```
ammen99:
```suggestion
			Whether to show a popup with the new values upon a change to the quick action target
```

wf-shell.ini.example
```diff
@@ -130,6 +130,37 @@ menu_logout_command = wayland-logout
 # after scrolling on the volume icon
 volume_display_timeout = 2.5
 
+# Configuration for the wireplumber mixer widget
```
ammen99, line 143: I would add a sentence (or a short paragraph) explaining what this widget does and how it compares to the volume widget. Gotta admit I had no idea what wireplumber is and what this widget was supposed to do before actually reviewing the code.

AKArien: Also changed the name to be more descriptive. If somebody is familiar with audio servers, they’ll probably get what « wp » is , if not, they’ll understand « mixer ».

AKArien:
@ammen99 This time should be final. Hopefully.
Updated the initial description again, but basically : fixed long names strecthing the widget, added tooltips, some margins and spacing, an option to put the icons on the left, and changed naming.
Also updated the icons, which applies back to the volume widget, OOR volume is now the emblem-unreadable icon.

Finally, i was wondering if you would like this widget to replace volume in the default configuration ?
With the adoption of pipewire out of the box on pretty much all major distros, it should work on the vast majority of systems, and i would like to think this widget is more polished than the pulseaudio one.
Against that stand the systems on which it will not work out of the box and that some of the functionalities might be confusing the manual by the user’s side. Assuming it all works flawlessly.

ammen99:
> Finally, i was wondering if you would like this widget to replace volume in the default configuration ?
> With the adoption of pipewire out of the box on pretty much all major distros, it should work on the vast majority of systems, and i would like to think this widget is more polished than the pulseaudio one.
> Against that stand the systems on which it will not work out of the box, how non-battle-tested it is, plus that some of the functionalities might be confusing the manual by the user’s side.

Would be best to first let the people test it for a while :) Otherwise I agree that this widget is in pretty much all ways superior to the old widget.

wf-shell.ini.example
```diff
@@ -125,11 +125,61 @@ menu_min_content_height = 500
...
+# This shows an almost full-featured mixer for the audio using Pipewire/Wireplumber
+# It allows for individual control of the volume of all apps, microphones and speakers,
+# as well as muting and choosing a default device for the latter two.
+# Supports scrolling and middle/right clicks with configurable actions (see below)
+# This default configuration shows off every feature
+
+# Which audio control will be the target of scrolling, muting... on the icon
+# Can be locked to the default input or output device
+wp_quick_target_choice = last_change # default_sink for output device, default_source for input
```
ammen99: Wouldn't default_sink be a better default btw? Especially since you mentioned how this could be a replacement for the default volume widget, I think usually volume widgets are supposed to handle the default sink.

metadata/panel.xml
```diff
@@ -249,6 +249,105 @@
 	</option>
 	</group>
 	<group>
+	<_short>Audio mixer (wireplumber)</_short>
+	<option name="wp_quick_target_choice" type="string">
+		<_short>Quick action target</_short>
+		<_long>Which audio control will be the target of the quick actions (scrolling and muting on the widget)</_long>
+		<default>last_change</default>
```
ammen99: Same as in the default ini, I think `default_sink` would be a better default. Otherwise nothing to say :)

wf-shell.ini.example
AKArien: Very true ! I guess i let myself get the better of me. will fix.

PR #311, opened 2025-11-11
panel: add Wi-Fi network menu with connection management

rummanz:
Implemented a Wi-Fi menu to the panel that allows users to view, select, and connect to wireless networks using nmcli. Includes password prompt, live status updates via DBus support.

### DIFF
--- a/src/panel/widgets/network.cpp
+++ b/src/panel/widgets/network.cpp
@@ -1,5 +1,4 @@
 #include "network.hpp"
-#include <atomic>
 #include <cassert>
 #include <gio/gio.h> // sometimes needed for lower-level variant helpers
 #include <giomm.h>
@@ -16,7 +15,6 @@
 #include <gdkmm/rgba.h>
 #include <glibmm/markup.h>
 #include <iostream>
-#include <mutex>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


rummanz:
I couldn't connect to any network using the dbus API. It shows the available networks but crashes while trying to connect to any SSID. 

here's the code block:
```void WayfireNetworkInfo::attempt_connect_ssid(const std::string &ssid,
  const std::string &password)
{
  if (ssid.empty() || ssid.find_first_not_of(" \t\r\n") == std::string::npos) {
      pop_status_label.set_text("Invalid SSID");
      return;
  }

  try {
      // Initialization (omitted for brevity, assume success)
      auto bus = Gio::DBus::Connection::get_sync(Gio::DBus::BusType::SYSTEM);
      auto nm = Gio::DBus::Proxy::create_sync(bus,
          "org.freedesktop.NetworkManager",
          "/org/freedesktop/NetworkManager",
          "org.freedesktop.NetworkManager");

      // --- Find Wi-Fi device (Mocked for compilation, assume real logic is used) ---
      // NOTE: In production code, you must replace this mock with the actual device detection.
      std::string wifi_device_path = "/org/freedesktop/NetworkManager/Devices/1"; 

      // Safety check for production device path detection
      if (wifi_device_path.empty()) {
          pop_status_label.set_text("No Wi-Fi device");
          return;
      }

      // --- Build settings with corrected dictionary structures (a{sv}) ---

      // 1. Create SSID byte array (type 'ay'). This is a full reference.
      std::vector<guchar> ssid_bytes(ssid.begin(), ssid.end());
      GVariant *ssid_array = g_variant_new_fixed_array(G_VARIANT_TYPE_BYTE,
                                               ssid_bytes.data(), ssid_bytes.size(), 1);

      // 2. Build 'connection' dictionary (a{sv})
      GVariantBuilder conn_builder;
      g_variant_builder_init(&conn_builder, G_VARIANT_TYPE("a{sv}"));

      g_variant_builder_add(&conn_builder, "{sv}", "type", g_variant_new_string("802-11-wireless"));
      g_variant_builder_add(&conn_builder, "{sv}", "id", g_variant_new_string(ssid.c_str()));
      gchar *uuid = g_uuid_string_random();
      g_variant_builder_add(&conn_builder, "{sv}", "uuid", g_variant_new_string(uuid));
      g_free(uuid);

      GVariant *connection_dict = g_variant_builder_end(&conn_builder); // Full reference

      // 3. Build '802-11-wireless' dictionary (a{sv})
      GVariantBuilder wifi_builder;
      g_variant_builder_init(&wifi_builder, G_VARIANT_TYPE("a{sv}"));
      
      // g_variant_new_variant consumes ssid_array (transfer-full).
      g_variant_builder_add(&wifi_builder, "{sv}", "ssid", g_variant_new_variant(ssid_array)); 
      g_variant_builder_add(&wifi_builder, "{sv}", "mode", g_variant_new_string("infrastructure"));

      GVariant *wifi_dict = g_variant_builder_end(&wifi_builder); // Full reference

      // 4. Build '802-11-wireless-security' dictionary (a{sv})
      GVariant *sec_dict = nullptr;
      if (!password.empty()) {
          GVariantBuilder sec_builder;
          g_variant_builder_init(&sec_builder, G_VARIANT_TYPE("a{sv}"));

          g_variant_builder_add(&sec_builder, "{sv}", "key-mgmt", g_variant_new_string("wpa-psk"));
          g_variant_builder_add(&sec_builder, "{sv}", "psk", g_variant_new_string(password.c_str()));

          sec_dict = g_variant_builder_end(&sec_builder); // Full reference
      }

      // 5. Build top-level array (a{sa{sv}}) - The entries for this array are of type {sa{sv}}

      // Create the entries using g_variant_new(), which creates a FLOATING reference 
      // and consumes the full references of the dictionary variants (connection_dict, etc.).
      GVariant *conn_entry = g_variant_new("{sa{sv}}", "connection", connection_dict);
      GVariant *wifi_entry = g_variant_new("{sa{sv}}", "802-11-wireless", wifi_dict);

      // Now add the floating entries to the top builder using g_variant_builder_add_value,
      // which consumes the floating reference. This is the safest way.
      GVariantBuilder top_builder;
      g_variant_builder_init(&top_builder, G_VARIANT_TYPE("a{sa{sv}}"));
      
      g_variant_builder_add_value(&top_builder, conn_entry);
      g_variant_builder_add_value(&top_builder, wifi_entry);
      
      if (sec_dict) {
          GVariant *sec_entry = g_variant_new("{sa{sv}}", "802-11-wireless-security", sec_dict);
          g_variant_builder_add_value(&top_builder, sec_entry);
      }

      GVariant *settings = g_variant_builder_end(&top_builder);
      
      if (!settings) {
          pop_status_label.set_text("Failed to build NetworkManager settings variant.");
          return;
      }

      // Glib::VariantContainerBase takes ownership (transfer-full)
      Glib::VariantContainerBase settings_wrap(settings, true);

      // --- Call AddAndActivateConnection (D-Bus signature: (a{sa{sv}}, o, o)) ---
      
      // Create the GVariant tuple: settings, device_path, connection_path
      // Use g_variant_ref() to increment the reference count of the existing settings variant 
      // so the new tuple construction can consume one reference.
      GVariant *args_final = g_variant_new("(@a{sa{sv}}oo)",
          g_variant_ref(settings_wrap.gobj()),
          wifi_device_path.c_str(), 
          "/"); 

      // Glib::VariantContainerBase takes ownership of the tuple (args_final).
      Glib::VariantContainerBase args_wrap(args_final, true);

      // --- Call AddAndActivateConnection ---
      nm->call_sync("AddAndActivateConnection", args_wrap);

      pop_status_label.set_text("Connecting...");
      update_active_connection();
  }
  catch (const Glib::Error &e) {
      std::cerr << "Connect failed: " << e.what() << std::endl;
      pop_status_label.set_text("Failed: " + Glib::ustring(e.what()));
  }
  catch (...) {
      pop_status_label.set_text("Unknown error");
  }
}```


And before crash it outputs these errors on console:
```(wf-panel:7112): Gtk-CRITICAL **: 01:18:19.327: gtk_box_append: assertion 'gtk_widget_get_parent (child) == NULL' failed

(wf-panel:7112): Gtk-WARNING **: 01:18:19.342: Theme parser error: <data>:1:28-33: Expected a valid color.
Font fallback .battery {font: 1rem default;}
Font fallback .clock {font: 1rem default;}

(wf-panel:7112): GLib-CRITICAL **: 01:18:27.500: g_variant_builder_end: assertion 'valid_builder' failed

(wf-panel:7112): GLib-CRITICAL **: 01:18:27.500: g_variant_get_type: assertion 'value != NULL' failed

(wf-panel:7112): GLib-CRITICAL **: 01:18:27.500: g_variant_type_is_array: assertion 'g_variant_type_check (type)' failed

(wf-panel:7112): GLib-CRITICAL **: 01:18:27.500: g_variant_get_type_string: assertion 'value != NULL' failed

(wf-panel:7112): GLib-ERROR **: 01:18:27.500: g_variant_new: expected array GVariantBuilder but the built value has type '(null)'
[1]    7112 trace trap (core dumped)  wf-panel
```

rummanz:
@ammen99 Thanks for your suggestion. Now it's working as expected. I need to enhance the UI a little bit then I will push the changes.

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,905 @@
...
+  auto header = Gtk::make_managed<Gtk::Label>("Available Wi-Fi networks");
+  header->set_margin_bottom(6);
+  pop_list_box.append(*header);
+
+  // Show scanning message
+  pop_status_label.set_text("Scanning for networks…");
+  pop_list_box.set_sensitive(false);
+
+  // Launch network scan in background
+  std::thread([this]()
```
ammen99: As far as I know, dbus also has an async interface. So I think it would be better to use the async interface instead of spawning a thread. The issue with the thread is that potentially there is a race condition where we start the scan and the widget is destroyed (granted that would be rare, maybe only if you disconnect the output while scanning?) but then I don't think we can easily stop the thread.

ammen99: And this would also remove the need for the signal_idle call I think.

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,905 @@
...
-    DBusProxy ap;
+  std::vector<NetworkInfo> result;
 
-    WifiConnectionInfo(const DBusConnection& connection, std::string path,
-        WayfireNetworkInfo *widget)
-    {
-        this->widget = widget;
+  try
+  {
+    auto connection = Gio::DBus::Connection::get_sync(Gio::DBus::BusType::SYSTEM);
```
ammen99: I am not sure how expensive this is but we already have a connection to dbus in the widget class, is there a specific reason why we reconnect here?

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,905 @@
...
 
-        return vstr.get();
-    }
+  return wifi.get() ? "wifi" : "network";
+}
 
-    std::string get_strength_str()
-    {
-        int value = get_strength();
+void WfNetworkConnectionInfo::spawn_control_center(DBusProxy &nm)
```
ammen99: I agree that left-click to show a list of wi-fi networks is way more useful than opening gnome-control-center, however, I think it would be good to still be able to spawn the control center (we could, in the future, update it with a custom command from the config file). Maybe we can do that on right click? So left click opens the list of wifi networks, right click spawns control center.

ammen99 (changes requested):
a few things I noticed on a first glance :)

src/panel/widgets/network.cpp
rummanz: Thanks for your valuable feedback. I have used dbus async interface and also removed the signal_idle call.

src/panel/widgets/network.cpp
rummanz: Thanks for identifying this issue. I think I have fixed this on latest commit.

src/panel/widgets/network.cpp
rummanz: You are right. It's good to have gnome-control-center as well as the popover menu. Latest commit now spawn the control center on right click and left click shows the popover menu.

ammen99:
@rummanz Formatting seems quite odd at times, most likely due to the high level of nested callbacks. Maybe we should split out some of the functionality into separate functions? I think that will improve readability as well. Also btw I don't think you need the connection to network manager on dbus to be async itself, but only the wifi scan, as that is what takes  a long time. The connection itself can be sync (like we already have), and you need to connect once at startup, so it is fine for that operation to be sync. This will also simplify the code a bit :)

rummanz:
@ammen99 Thanks for your valuable feedback. I have refactored the code and removed async call to network manager dbus api. Only network scan is now async.

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,1216 @@
...
+  int get_strength() {
+    assert(ap);
+
+    Glib::Variant<guchar> vstr;
+    ap->get_cached_property(vstr, STRENGTH);
+
+    return vstr.get();
+  }
+
+  std::string get_strength_str() {
```
ammen99: here we have the same thing as above, I feel like all of this should be one function returning whichever properties we want, not multiple functions with the same if cascade.

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,1216 @@
...
-    command += get_control_center_section(nm);
-
-    Glib::spawn_command_line_async(command);
+struct NetworkInfo {
+  std::string ssid;
+  std::string path;
+  int strength = 0;     // Wi-Fi signal strength (0–100)
+  bool secured = false; // True if the network requires authentication
+};
+// Map 0-100 strength to GNOME symbolic icon names
```
ammen99: Seems to me like we have a different gradient (here 5-30-55-80, below for the colors 0-25-40-100), maybe we should put both in the same list (i.e change `status_color` so that it includes the icon name as well, then have one function to compute both?)

I think the real issue here is theming, in the gtk4 port (cc @trigg) the color gradient was removed, but it seems we forgot to remove the code for it. Now you have fixed it, but I suppose that the css properties won't work anymore (seems like you removed the classes in this PR?): https://github.com/WayfireWM/wf-shell/blob/57a68cb42c8b1a4a5a3f5163e3ca0899804bead9/src/panel/widgets/network.cpp#L274-L289

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,1216 @@
...
+static std::string icon_name_for_strength(int value) {
+  if (value > 80)
+    return "network-wireless-signal-excellent-symbolic";
+  if (value > 55)
+    return "network-wireless-signal-good-symbolic";
+  if (value > 30)
+    return "network-wireless-signal-ok-symbolic";
+  if (value > 5)
+    return "network-wireless-signal-weak-symbolic";
+  return "network-wireless-signal-none-symbolic";
```
ammen99, line 55: This logic is also duplicated in `get_icon_name()` in the wifi connection type implementation.

ammen99 (changes requested):
I tested this very briefly on my laptop, some things I noticed:

- The widget, in 'idle' state (i.e not showing menu or anything) does not seem to display the output of the active wifi connection, as the current implementation in master does. I am not sure why this part was removed from the PR.

- When using left-click, it triggers a scan and then the networks are populated. I think we can cache the old networks (from a previous scan) and then update the list once we get the new results. Maybe depending on how old the previous scan was. I mostly mean it for cases where you click, then dismiss, then immediately open the menu again, it looks weird that we have to wait for the results to open up. At the very least, I think while we are scanning, we should display the currently connected network. 

- I seem to be getting my Wi-Fi network twice, once as the current connection, once as an option to connect to it, that is probably not what we want?

- The popover seems to be bigger than it needs to be to accomodate the list of networks, I am not sure whether there is anything we can do about it.

rummanz:
@ammen99  Thanks for testing and reviewing the code. I have tried to address some of the issues in the latest commit.

> * The widget, in 'idle' state (i.e not showing menu or anything) does not seem to display the output of the active wifi connection, as the current implementation in master does. I am not sure why this part was removed from the PR.

I have removed this function intentionally to make it more aesthetically pleasing but I guess that's not what most people want. So I have added it back.

> * When using left-click, it triggers a scan and then the networks are populated. I think we can cache the old networks (from a previous scan) and then update the list once we get the new results. Maybe depending on how old the previous scan was. I mostly mean it for cases where you click, then dismiss, then immediately open the menu again, it looks weird that we have to wait for the results to open up. At the very least, I think while we are scanning, we should display the currently connected network.

That's a good idea and I think that's how it should work but unfortunately I have no idea how to achieve that, I tried to cache the last scan but then again it starts to show the entire list of SSIDs twice including connected network and also it can't sort networks by strength. However, right now It shows the connected network while scanning.

> * I seem to be getting my Wi-Fi network twice, once as the current connection, once as an option to connect to it, that is probably not what we want?

This is an interesting one because I did not face this issue while testing on my laptop probably because I was using iwd as wifi backend. So when I switched to wpa_supplicant, I saw there's a duplicate entry of connected network. I can confirm that latest commit fixed this issue with wpa_supplicant. 

> * The popover seems to be bigger than it needs to be to accomodate the list of networks, I am not sure whether there is anything we can do about it.

In my laptop it shows smaller than the notification popup menu. I can make it narrower but then some long SSID won't show properly. I think we can resize the popover menu if you can suggest me an ideal size.

src/panel/widgets/network.cpp
```diff
@@ -1,458 +1,1286 @@
...
-
-    return wifi.get() ? "wifi" : "network";
+struct NetworkInfo {
+  std::string ssid;
+  std::string path;
+  int strength = 0;     // Wi-Fi signal strength (0–100)
+  bool secured = false; // True if the network requires authentication
+};
+// Unified strength bucketing (0-100) used for labels and icons
+static const char *strength_bucket(int value) {
```
NamorNiradnug, line 36: This could return `std::string_view`

trigg:
Closing after merging #333 

Thank you for the PR

PR #321, opened 2025-12-30
Allow accessing alternative clicks via touchscreen long press

trigg:
Adds the following long press gestures to access alternative click actions:

- Extra actions on application menu launchers
- Window list menu
- Volume control mute
- Notifications D&D
- Tray items menu

Also splits the logic of click press & click release to allow these events to be cancelled correctly.

### DIFF
--- a/data/css/default.css
+++ b/data/css/default.css
@@ -98,27 +98,37 @@
 }
 
 .wf-dock .box {
+    background-color: #33333355;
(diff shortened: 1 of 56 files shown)


trigg:
For rationale, on a touch-only device these features are completely impossible to reach in current release. Not major blockers but it would make sense to bring them across.

PR #322, opened 2026-01-02
Improve dock

soreau:


src/dock/toplevel-icon.cpp
```diff
@@ -147,19 +147,10 @@ class WfToplevelIcon::impl
...
-        bool is_activated = this->state & WF_TOPLEVEL_STATE_ACTIVATED;
         bool is_min = state & WF_TOPLEVEL_STATE_MINIMIZED;
         bool is_max = state & WF_TOPLEVEL_STATE_MAXIMIZED;
         auto style  = this->button.get_style_context();
-        if (!was_activated && is_activated)
-        {
-            style->remove_class("flat");
-        } else if (was_activated && !is_activated)
-        {
-            style->add_class("flat");
```
trigg, line 161: Removing flat is a good idea, but I figure we probably need an 'activated' class for is_activated still

```
if (is_activated)
{
    style->add_class("activated");
} else
{
    style->remove_class("activated");
}
```
This will allow CSS writers to differentiate activated window

soreau:
@trigg I'm done fiddling with the activated css, it just renders a white line beneath the active button icon.

PR #326, opened 2026-01-19
Fix memory leaks ; remove useless headers ; use pragma once ; renames of variables herited from gtk3 ; removed unused stuff

AKArien:
Fixes memory leaks in the various widgets. Mostly happened when reloading the configuration by not disconnecting signals, but in the case of the menu widget, there were also leaks for each show/hide.

For the rest that is on this branch, i’m fully prepared to hear it’s out of the scope of this pr, but oh well, that’s what git cherry-pick is for :
- #pragma once instead of ifdefs
- remove pointless includes
- use <> for includes from libraries and "" for wf-shell headers
- split WayfireMenu::update_popover_layout into it and setup_popover_layout for clarity
- change hbox to box in WayfireStatusNotifier

### DIFF
--- a/src/dock/dock.hpp
+++ b/src/dock/dock.hpp
@@ -1,12 +1,11 @@
-#ifndef WF_DOCK_HPP
-#define WF_DOCK_HPP
+#pragma once
 
 #include <gtkmm/box.h>
 #include <wayland-client.h>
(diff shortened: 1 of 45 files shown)


trigg:
Overall looks very good. Ran and did some testing and see no negatives.

Tray in its current incarnation is an abomination I designed with no previous understanding of dbus, I'm not too surprised it has leaks.

PR #327, opened 2026-01-21
panel : brightness control widget

AKArien:
Add a brightness widget to control display brightness.

Handles brightness for displays exposed in sysfs, and external external monitors that implement MCCS. MCCS control depends on libddcutil, and is not built if it is not detected.
Presents a slider per supported device in a popover on click. If we can detect which display a device corresponds to, we isolate them in the view of the panel of this display, and scrolling on the widget changes it.
Has the classic niceties of scrolling and pop out on detected change, although we don’t have a way to watch for changes on MCCS displays.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - run: echo 'http://dl-cdn.alpinelinux.org/alpine/v3.22/community' > /etc/apk/repositories
       - run: echo 'http://dl-cdn.alpinelinux.org/alpine/v3.22/main' >> /etc/apk/repositories
-      - run: apk --no-cache add git g++ binutils pkgconf meson ninja musl-dev gtkmm4-dev vala gobject-introspection gobject-introspection-dev pulseaudio-dev pipewire-dev wireplumber-dev libdbusmenu-glib-dev alsa-lib-dev yyjson-dev linux-pam-dev util-linux-login openssl-dev
+      - run: apk --no-cache add git g++ binutils pkgconf meson ninja musl-dev gtkmm4-dev vala gobject-introspection gobject-introspection-dev ddcutil pulseaudio-dev pipewire-dev wireplumber-dev libdbusmenu-glib-dev alsa-lib-dev yyjson-dev linux-pam-dev util-linux-login openssl-dev
       - run: echo 'http://dl-cdn.alpinelinux.org/alpine/edge/testing' >> /etc/apk/repositories
       - run: echo 'http://dl-cdn.alpinelinux.org/alpine/edge/main' >> /etc/apk/repositories
       - run: echo 'http://dl-cdn.alpinelinux.org/alpine/edge/community' >> /etc/apk/repositories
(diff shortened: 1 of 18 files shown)


trigg:
```
diff --git a/src/panel/widgets/backlight/sysfs.cpp b/src/panel/widgets/backlight/sysfs.cpp
index 2db42c2..5dcddac 100644
--- a/src/panel/widgets/backlight/sysfs.cpp
+++ b/src/panel/widgets/backlight/sysfs.cpp
@@ -133,6 +133,7 @@ class WfLightSysfsDbusSystemdControl : public WfLightSysfsControl
 
     void set_brightness(double brightness) override
     {
+        brightness = std::clamp(brightness, 0.0, 1.0);
         auto params = Glib::VariantContainerBase::create_tuple({
             Glib::Variant<Glib::ustring>::create("backlight"),
             Glib::Variant<Glib::ustring>::create(name),
```

A fix for the out-of-range GDBUS errors

src/util/icons.cpp
```diff
@@ -0,0 +1,19 @@
+#include "icons.hpp"
+
+std::string icon_for(std::map<double, std::string> icons, double value)
+{
+    if ((value > ICONS_VMAX) || (value < ICONS_VMIN))
```
trigg: I'll need to test this specifically, but be aware volume is quite normal to go > 100% via pavucontrol and other methods.

I'd advise putting a `audio-volume-high-danger-symbolic` at >100% and clamping the top end to not show the error icon

trigg (changes requested):

src/util/icons.cpp
trigg: `pactl set-sink-volume @DEFAULT_SINK@ 120%`

And mind your ears.

AKArien: sounds like a good way to handle it for now. i do want to implement a multiplier for the mixer eventually, but that’s eventually.

trigg (approved):
Testing comes up good on my devices, code looks good too!

AKArien:
> Testing comes up good on my devices, code looks good too!

glad you think so ! but it’s still in draft for a few reasons :)

With a certain monitor of mine, it reliably crashes somewhere in libddcutil, so i’ll try to see if i can at least not make it so. Also tends to crash with suspend rather randomly, so i think i will just use the gdk events rather than libddcutil’s watch.
There’s also the general layout to make better and add style classes.
I also wanted to add support for more generic leds, but i think that will be for a later update.

trigg:
As in chat, but for the record.

All my usb hubs break DDC/CI support to monitor. Desktop to monitor and even Desktop via KVM to monitor works fine.

Other than a small tweak which I've commited now, I'd say this looks ready!

I'm not sure if 'omit entirely' or 'show but deactivated' is better for monitors which lack DDC/CI support, but as we have already arrived at 'omit entirely' I think that is acceptable.

AKArien:
@trigg thanks for the help !!
Things seem to work very good now. I’ll try whacky-as-possible things to check if it does work as expected, but i’ll probably undraft, count it as ready and leave matters to your own hands tomorrow :)

Issue #269, opened 2024-08-31
panel: harmonise icon sizing


headbank:
Just been making first steps in wayfire and one thing that struck me was how many individual steps are needed to resize the panel consistently. You need to change the icon-size for every plugin that has one (including some that are not mentioned in the wf-shell.ini.example document). Even using wcm this is a chore.

I propose that there be a global setting for icon size, that the individual plugin settings can override if specified.

ammen99:
This might be possible with CSS, but I am no expert.

cc @trigg ?

trigg:
Changing Icon size by CSS is possible in [GTK4](https://docs.gtk.org/gtk4/css-properties.html#gtk-css-properties), but not 3.

headbank:
Perhaps I'm misconstruing the intent of the comments above, but they seem to pertain to users being able to hack on CSS files to achieve default size, which gets away from the kind of usability enhancement I had in mind, namely a global default option in `wf-shell.ini`. (Perhaps you are in fact discussing internals of how to implement it, in which case apologies for the misunderstanding.)

There must already be a coded default size that applies when no `*_icon_size` value has been set for a plugin in `wf-shell.ini`. My thought was that this could be exposed as an option under `[panel]` but that the existing per-plugin values could still override it if specified.

Additional thought: If user does not specify a global default size, maybe it would make some sense for it to be dynamically derived from the `minimal_height` value?

mrfoggg:
I tried to reduce the height of the panel - I reduced all the icons, set min-height to 1, but the panel is still too high and a disproportionate indent appeared at the bottom

AKArien:
Can be closed as of #328

PR #328, opened 2026-01-22
panel: unify icon sizes

AKArien:
By default, icons of widgets will now follow the size of the panel.
Adds a css config for a default size for icons, which is overriden by the configuration entry for each icon if it is not 0. Meanwhile, the default becomes 0.
Adjusted the panel’s default height to keep icon sizes the same as the default configuration (32), and left the menu and launchers at 42. Personally, i would think that setting them all to 0 and setting the panel’s default minimal height to 42 would be preferable, it would also lead to other widgets being slightly bigger by default.

The icon size of launchers was getting applied to the menu widget as well, probably because it didn’t have a config option of it’s own. Added one.
This also fixes command-output’s size not actually being accounted for.

Example/default ini file has not yet been changed for any of this.

Also, potential breaking change we could for the sake for name consistency, if we value this more than not breaking things (probably not very wise but it annoys me personally) : rename laucher’s icon size `launchers_size` to `launchers_icon_size`.

Addresses #269

### DIFF
--- a/src/panel/widgets/command-output.cpp
+++ b/src/panel/widgets/command-output.cpp
@@ -57,11 +57,11 @@ WfCommandOutputButtons::CommandOutput::CommandOutput(const std::string & name,
 
     if (icon_size > 0)
     {
-        auto provider = Gtk::CssProvider::create();
-        provider->load_from_string(".command-icon-" + name + "{-gtk-icon-size:" + std::to_string(
+        css_provider = Gtk::CssProvider::create();
+        css_provider->load_from_string(".command-icon-" + name + "{-gtk-icon-size:" + std::to_string(
             icon_size) + "px;}");
         icon.get_style_context()->add_class("command-icon-" + name);
-        icon.get_style_context()->add_provider(provider, GTK_STYLE_PROVIDER_PRIORITY_USER);
+        icon.get_style_context()->add_provider(css_provider, GTK_STYLE_PROVIDER_PRIORITY_USER);
     }
 
     icon.get_style_context()->add_class("widget-icon");
--- a/src/panel/widgets/command-output.hpp
+++ b/src/panel/widgets/command-output.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "../widget.hpp"
+#include "glibmm/refptr.h"
 
 #include <gtkmm/button.h>
 #include <gtkmm/image.h>
@@ -21,6 +22,8 @@ class WfCommandOutputButtons : public WayfireWidget
         Gtk::Box box;
         Gtk::Image icon;
         Gtk::Label main_label;
+
+        Glib::RefPtr<Gtk::CssProvider> css_provider;
 
         Gtk::Label tooltip_label;
         time_t last_tooltip_update = 0;


src/panel/panel.cpp
```diff
@@ -349,28 +349,63 @@ class WayfirePanelApp::impl
...
 void WayfirePanelApp::on_config_reload()
 {
+    reload_css();
     for (auto& p : priv->panels)
     {
         p.second->handle_config_reload();
     }
 }
 
+void WayfirePanelApp::reload_css()
```
trigg: The CssFromConfig classes have their own reload logic and shouldn't be made/destroyed on config changes. 

That said, one extra Int variant for Icons would be a good idea, where 0 explicitly sets an empty string for style, allowing CSS to fall back to the default icon rule

src/panel/widgets/command-output.cpp
```diff
@@ -50,9 +52,19 @@ WfCommandOutputButtons::CommandOutput::CommandOutput(const std::string & name,
...
+
     if (icon_size > 0)
     {
-        image_set_icon(&icon, icon_name);
+        auto provider = Gtk::CssProvider::create();
+        provider->load_from_string(".command-icon-" + name + "{-gtk-icon-size:" + std::to_string(
+            icon_size) + "px;}");
+        icon.get_style_context()->add_class("command-icon-" + name);
+        icon.get_style_context()->add_provider(provider, 1);
+    } else
```
trigg: This else needs to be flattened to make sure "widget-icon" is always set on command-output icons

src/util/css-config.cpp
```diff
@@ -45,6 +45,27 @@ CssFromConfigInt::CssFromConfigInt(std::string option_name, std::string css_befo
     add_provider();
 }
 
+CssFromConfigIconSize::CssFromConfigIconSize(std::string option_name, std::string css_class) : option_value{option_name}
+{
+    provider = Gtk::CssProvider::create();
+
+    auto reload_css = [=]()
```
trigg: Good idea, using a named lambda. Completely missed that possibility myself!

src/panel/widgets/notifications/notification-center.cpp
```diff
@@ -10,7 +10,8 @@
 void WayfireNotificationCenter::init(Gtk::Box *container)
 {
     button = std::make_unique<WayfireMenuButton>("panel");
-    button->get_style_context()->add_class("notification-center");
```
trigg, line 13: The 'notification-center' class should be on the top-most widget here, otherwise we'd have no way to affect it with a CSS rule. It should be removed from the icon.

src/panel/panel.cpp
```diff
@@ -370,14 +370,25 @@ void WayfirePanelApp::on_config_reload()
...
+    const static std::vector<std::pair<std::string, std::string>> icon_sizes_args =
+    {
+        {"panel/minimal_height", ""},
+        {"panel/menu_icon_size", ".menu-icon"},
+        {"panel/launchers_size", ".launcher"},
+        {"panel/battery_icon_size", ".battery image"},
+        {"panel/network_icon_size", ".network"},
+        {"panel/volume_icon_size", ".volume"},
+        {"panel/wp_icon_size", ".wireplumber"},
+        {"panel/notifications_icon_size", ".notification-center"},
```
trigg: As above I've mentioned moving .notification-center back to the parent widget, this line needs to be `".notification-center "` to account for the change

src/panel/widgets/command-output.cpp
AKArien: oops ! remnant from my first try

src/util/css-config.cpp
AKArien: to take into your cssfromconfig deduplication :)

trigg (commented):
LGTM, will do a trial run shortly.

PR #329, opened 2026-01-26
panel: Add weather widget

soreau:
Depends on[ open-weather-fetch](https://github.com/Secret-Chest/open-weather-fetch). Use `-Dweather=true` to enable.

### DIFF
--- a/src/panel/widgets/weather.cpp
+++ b/src/panel/widgets/weather.cpp
@@ -30,12 +30,12 @@ void WayfireWeather::init(Gtk::Box *container)
 
 bool WayfireWeather::handle_inotify_event(Glib::IOCondition cond)
 {
-	if (cond == Glib::IOCondition::IO_HUP)
-	{
+    if (cond == Glib::IOCondition::IO_HUP)
+    {
         return false;
     }
 
-	char buf[1024 * sizeof(inotify_event)];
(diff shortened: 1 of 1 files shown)


src/panel/widgets/weather.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+    label.set_justify(Gtk::Justification::CENTER);
+    box.get_style_context()->add_class("weather");
+    box.prepend(label);
+    box.append(icon);
+    container->append(box);
+
+    update_weather();
+
+    timeout = Glib::signal_timeout().connect_seconds(
```
trigg: Not using file watching? :D

trigg (approved):
LGTM!

src/panel/widgets/weather.cpp
soreau: Probably not a bad idea..

soreau: Does it work for symlinks though if the symlinks doesn't change but the file pointed to does?

soreau: I just looked it up, apparently it follows the symlink to the actual file unless `IN_DONT_FOLLOW` is passed to `inotify_add_watch()`.

PR #333, opened 2026-02-03
util: network manager

trigg:
Wrote a central util for tracking network state.

On the base level you `NetworkManager::getInstance()` to start using it.

With this, we can share a broad-featured network control panel between panel and dock, and allow incoming locker to track state in a familiar way.

The network widget on the panel has only minor changes over all:
- include "with ${VPN_NAME}" in name when a VPN is connected
- clicking it no longer runs a command, it now has an in-built popover with custom controls

Features required before considered ready:
- Right click / Long touch to open settings/gnome cc


hustlerone:
i would love to see this to be honest

trigg:
~Needs a fix around leftover widgets when ModemManager stops/starts~

Otherwise dog fooding is showing positive.

trigg:
modem manager quirk has been fixed.

New issues:
~- Creating or Removing a VPN via nmtui or nm-connection-editor does not show up until next run~

trigg:
The `-secure` variants of icons are not present in many tested themes. I'll have to re-plan the widget with an overlay and a separate VPN `secure` icon

PR #335, opened 2026-02-10
Apply the flat style to command-output

AKArien:


trigg (approved):
LGTM! thanks for the fix

PR #337, opened 2026-02-13
fix error handling in ipc client and add safety checks

AKArien:


src/panel/widgets/language.cpp
```diff
@@ -25,6 +26,13 @@ void WayfireLanguage::init(Gtk::Box *container)
     ipc_client->subscribe(this, {"keyboard-modifier-state-changed"});
     ipc_client->send("{\"method\":\"wayfire/get-keyboard-state\"}", [=] (wf::json_t data)
     {
+        if (data.serialize().find(
+            "error") != std::string::npos)
```
soreau, line 30: Did uncrustify separate this into two lines? Doesn't seem necessary..

src/util/wf-ipc.cpp
```diff
@@ -30,34 +31,54 @@
...
+        connection = client->connect(address);
+        connection->get_socket()->set_blocking(false);
+        output = connection->get_output_stream();
+        input  = connection->get_input_stream();
+        cancel = Gio::Cancellable::create();
+
+        return true;
+    } catch (const Glib::Error& ex)
+    {
+        std::cerr << "Error connecting to WAYFIRE_SOCKET: " << ex.what();
```
soreau: Hm, should we output the `socket_path` here?

PR #339, opened 2026-02-17
panel: Add workspace-switcher widget

soreau:
Requires #340 for proper expected functionality.

### DIFF
--- a/src/panel/widgets/workspace-switcher.cpp
+++ b/src/panel/widgets/workspace-switcher.cpp
@@ -355,11 +355,11 @@ void WayfireWorkspaceSwitcher::process_workspaces(wf::json_t workspace_data)
                 return;
             }
 
-            auto output_id     = output_data["id"].as_int();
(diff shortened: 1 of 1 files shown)


AKArien:
tested quickly a bit, and i notice, if i switch the workspace row by anything else than the widget :
- if i changed workspace row, switching with the widget takes me back the previous row i was on last time i switched vertically by the widget (by scrolling)
- the workspace i am on will be at the center, and the widget’s representation of the surrounding workspaces will be positionned relative to the current workspace

soreau:
> tested quickly a bit, and i notice, if i switch the workspace row by anything else than the widget :
> 
>     * if i changed workspace row, switching with the widget takes me back the previous row i was on last time i switched vertically by the widget (by scrolling)
> 
>     * the workspace i am on will be at the center, and the widget’s representation of the surrounding workspaces will be positionned relative to the current workspace

Thanks, I'd noticed this too but forgot about it. Should be fixed now.

trigg:
Looking good, ready to merge?

PR #343, opened 2026-02-19
Switcher : use WfMenuButton ; harmonise size ; add grid mode ;  adjust naming

AKArien:
Cleaned commit history, should be good as is if all changes are welcome

Using a WayfireMenuButton for the popover mode allows keeping the panel out when the menu is open and prevents issues ranging from uglyness to unusability.

Like other icons, the switcher now can be made as big as the panel by setting size as 0, and this is the new default.

Added a grid mode which displays the grid from the popover directly in the panel. Size is recalculated so it satisfies the size configuration as a whole.

Changed names of the existing mode options to « row » and « grid_popover ».

Updated naming in the code to reflect these changes and be clearer, and use option.value().

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -472,16 +472,16 @@ Set to -1 to only run it by clicking the button.
 	<option name="workspace_switcher_mode" type="string">
 		<_short>Workspace switcher widget mode</_short>
 		<desc>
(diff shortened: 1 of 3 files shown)


trigg:
Thank you for the hard work :D

PR #349, opened 2026-02-20
fix pipewire/wireplumber dependencies in build

AKArien:
Not getting the pipewire dependency and relying on the wireplumber one breaks on some systems.

Additionally, change the option name from wireplumber to pipewire.


meson_options.txt
```diff
@@ -5,10 +5,10 @@ option(
     description: 'Build pulseaudio volume widget',
 )
 option(
-    'wireplumber',
+    'pipewire',
```
soreau: I would rename this option to wp_mixer.

AKArien: The thing is, the option with the pulse-audio volume widget isn’t named volume widget either, it’s simply pulse. For consistency, i’d want to rename both atp

soreau: Yes, I agree this would be an improvement.

@trigg What are your thoughts?

trigg: Seems like a good plan. Name options for the feature they add, not the library they pull in

PR #353, opened 2026-02-21
background: Don't inhibit

soreau:
Wayfire now handles this internally.

Reference: https://github.com/WayfireWM/wayfire/pull/2978


trigg:
Great! thanks again

PR #355, opened 2026-02-21
Updated background_color examples to match GTK4 css style format

mrsteve0924:
i updated the wiki
https://github.com/WayfireWM/wf-shell/wiki/wf%E2%80%90panel#background_color

### DIFF
--- a/wf-shell.ini.example
+++ b/wf-shell.ini.example
@@ -40,10 +40,11 @@ layer = top
 # Possible values are only "top" and "bottom"
 position = top
 
-# set the background color.
+# set the background color
 background_color = gtk_headerbar # match the color of a GtkHeaderbar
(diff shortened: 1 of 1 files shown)


trigg:
Good idea to update the docs, but there could be a fair bit more information here. I don't know how much we'd want to cover manually and how much we could point to outside docs for.

https://www.w3.org/TR/css-color-5/

essentially:
- Three digit hex codes : #RGB
- Four : #RGBA
- Six : #RRGGBB
- Eight : # RRGGBBAA
- rgb() / rgba()
- hsl() / hsla()
- hwb()
- named colors (red, dimgray, coral, wheat)

added: also gtk_headerbar is now non-functional and just uses window colour, as there was no way to poll current theme colours out that I could find

trigg:
Thanks for the help !

PR #356, opened 2026-02-21
panel: Add outputs option for per-output configuration

soreau:
This adds a new option named 'outputs' that is a list of output names on which a panel instance should be rendered. The default value is '*' wildcard, which means all outputs.

### DIFF
--- a/src/panel/panel.cpp
+++ b/src/panel/panel.cpp
@@ -442,23 +442,16 @@ bool WayfirePanelApp::panel_allowed_by_config(bool allowed, std::string output_n
 
 void WayfirePanelApp::update_panels()
 {
-    /* First case: Panel exists on the output but is not allowed by config: Remove panel */
     for (auto& o : *get_wayfire_outputs())
     {
         auto output = o.get();
         auto output_name = o->monitor->get_connector();
+
         if (panel_allowed_by_config(false, output_name))
         {
             std::cout << "Removing panel from output: " << output_name << std::endl;
             priv->panels.erase(output);
         }
-    }
-
-    /* Second case: Panel does not exist for the output but is allowed by config: Add panel */
-    for (auto& o : *get_wayfire_outputs())
-    {
-        auto output = o.get();
-        auto output_name = o->monitor->get_connector();
 
         const auto it = std::find_if(priv->panels.begin(), priv->panels.end(),
             [&output_name] (const auto& panel)


src/panel/panel.cpp
```diff
@@ -411,6 +411,8 @@ class WayfirePanelApp::impl
 {
   public:
     std::map<WayfireOutput*, std::unique_ptr<WayfirePanel>> panels;
+    std::map<std::string, WayfireOutput*> outputs;
+    WfOption<std::string> *panel_outputs;
```
ammen99: Having this dynamically allocated seems really weird, if you need to load an option at runtime, you can use `.load_option(<option name>)`

src/panel/panel.cpp
```diff
@@ -425,6 +427,73 @@ void WayfirePanelApp::on_activate()
 {
     WayfireShellApp::on_activate();
 
+    priv->panel_outputs->set_callback([=] ()
+    {
+        /* First case: Add a panel for each output that does not have one,
+         * because an '*' character was found in the panel_outputs string */
+        if (std::string(*priv->panel_outputs).find("*") != std::string::npos)
+        {
+            for (auto& o : priv->outputs)
```
ammen99: Why do we need a separate map from string to a WayfireOutput ? If we want to iterate over all outputs, I believe gdk has a function for this (https://docs.gtk.org/gdk4/method.Display.get_monitors.html)

src/panel/panel.cpp
```diff
@@ -425,6 +427,73 @@ void WayfirePanelApp::on_activate()
...
+                    priv->panels[o.second]->handle_config_reload();
+                    priv->panels[o.second]->set_panel_app(this);
+                    priv->panels[o.second]->init_widgets();
+                }
+            }
+
+            return;
+        }
+
+        /* Second case: Add a panel for each output in the panel_outputs string,
```
ammen99: I believe a simpler algorithm would be like this in pseudocode:

```
for_each_existing_panel:
    if output not allowed in config:
        remove panel

for_each_gdk_monitor:
    if output allowed in config and does not have panel:
        create panel
```

src/panel/panel.cpp
```diff
@@ -466,8 +535,15 @@ std::shared_ptr<WayfireIPC> WayfirePanelApp::get_ipc_server_instance()
 
 void WayfirePanelApp::handle_new_output(WayfireOutput *output)
 {
-    priv->panels[output] = std::unique_ptr<WayfirePanel>(
-        new WayfirePanel(output));
+    priv->panel_outputs = new WfOption<std::string>{"panel/outputs"};
+    priv->outputs[output->monitor->get_connector()] = output;
+    if ((std::string(*priv->panel_outputs).find("*") != std::string::npos) ||
```
ammen99: We have these checks everywhere, I think it would make sense to add a function `is_panel_enabled_in_config(config_opt, output_name)`

src/panel/panel.cpp
ammen99: This way I believe you won't need any of the new maps you have added, and I have found that generally the less state you have, the less possibilities for things to go wrong.

src/panel/panel.cpp
soreau: Sounds like this might work.

src/panel/panel.cpp
soreau: Yes, you need the WayfireOutput to create the panel. You cannot get this from the gdk function, unless you make the wf-shell-app monitors list public, as far as I understand.

src/panel/panel.cpp
soreau: I think the pseudo code you have is flawed simply because we do not want to unnecessarily recreate panels, since that will reorder the window-list and reset other widgets.

src/panel/panel.cpp
soreau: That might be doable, but it's really only two cases where we'd need it, if I am thinking correctly.

src/panel/panel.cpp
ammen99: Sounds like we should add a getter then. The less overall state, the better IMO. Considering we already keep it.

soreau: A getter would incur looping each time we call the function and would require modifying code outside of panel.cpp, if I am thinking about this right. A std::map is more efficient here and keeps the patch clean.

src/panel/panel.cpp
ammen99: I don't see how it is flawed, because I say to create a panel if it does NOT exist already.

soreau: Well it's not very different from what I have so far, I'm sure there are many other ways to skin this cat though.

trigg:
Sorry, I've read through it.

I currently don't know where I stand on the conversation but wonder why you're specifically avoiding doing this in shell-app.

soreau:
> ... but wonder why you're specifically avoiding doing this in shell-app.

@trigg Thanks for looking. It is not the case that I set out with the intent to only touch panel.cpp, it's just the way the patch turned out. After I had it working, I noticed the only thing I had in wf-shell-app.hpp was moving the monitors list from private to public, and I never used it, in favor of the more convenient std::map that I'd installed.

I will have another look to see if using the existing monitors list is doable and get rid of the extra std::map though, if that's what you're expecting.

src/panel/panel.cpp
soreau: Well this sounded good but I was not able to get `load_option()` working to help here. Maybe I am using it wrong.

src/panel/panel.cpp
```diff
@@ -411,20 +411,116 @@ class WayfirePanelApp::impl
...
+}
+
+void WayfirePanelApp::update_panels()
+{
+    /* First case: Panel exists on the output but is not allowed by config: Remove panel */
+    for (auto& o : *get_wayfire_outputs())
+    {
+        auto output = o.get();
+        auto output_name = o->monitor->get_connector();
+        if (panel_allowed_by_config(false, output_name))
```
trigg: Any reason this can't be one loop? Unless I'm misunderstanding this could be done at the top of Second case loop.

trigg, line 450: Any reason this can't be one loop? Unless I'm misunderstanding this could be done at the top of Second case loop.

trigg (commented):
My read through looks like the loops can be merged. Otherwise very happy with this, and if you're sure of yourself just merge this :)

src/panel/panel.cpp
soreau: Nope, no reason. Good eyes! I don't know why I thought two loops were needed. 😀

PR #358, opened 2026-02-22
panel : menu : fix label ; fix extra actions for grid view ; added css sizing/styling

AKArien:
Extra actions were previously made unavailable for grid mode, added them back.

The label position was put always on the side instead of the bottom for grid mode, put it back below.

General cleanups of the implementation and declaration of WfMenuMenuItem, as well as renamed it to WfMenuItem.

Made menu items sizeable by css (app-button) and by a new config option.

New configuration option to set spacing between menu items.

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -59,69 +59,130 @@ void WfMenuCategoryButton::on_click()
     menu->set_category(category);
 }
 
-WfMenuMenuItem::WfMenuMenuItem(WayfireMenu *_menu, Glib::RefPtr<Gio::DesktopAppInfo> app) :
-    Gtk::FlowBoxChild(), menu(_menu), m_app_info(app)
+WfMenuItem::WfMenuItem(WayfireMenu *_menu, Glib::RefPtr<Gio::DesktopAppInfo> app) :
+    Gtk::FlowBoxChild(), menu(_menu), app_info(app)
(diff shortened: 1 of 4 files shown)


AKArien:
@trigg you’ll have to tell me if you too prefer this set of bugs ;)

trigg:
All the context-menu options are inactive in this PR.

src/panel/widgets/menu.cpp
```diff
@@ -59,69 +59,130 @@ void WfMenuCategoryButton::on_click()
...
+        extra_actions_button.activate();
+        right_click_g->set_state(Gtk::EventSequenceState::CLAIMED);
+    }));
+    signals.push_back(long_press_g->signal_pressed().connect(
+        [=] (double x, double y)
+    {
+        extra_actions_button.activate();
+        long_press_g->set_state(Gtk::EventSequenceState::CLAIMED);
+        left_click_g->set_state(Gtk::EventSequenceState::DENIED);
+        right_click_g->set_state(Gtk::EventSequenceState::DENIED);
```
trigg: No need to deny right_click_g here as it can't have been started. I don't think there's ramifications for doing this, but it isn't necessary.

A left click gesture would have started on a touch, but not a ->set_button(3) gesture

AKArien: Oops !

PR #360, opened 2026-02-22
Make menu layer override when showing popup optional

AKArien:
Currently, when the menu widget pops out it’s popover, it sets the panel layer to GTK_LAYER_SHELL_LAYER_OVERLAY. I am pretty sure it doesn’t really have a place here, as a user desiring this would just set the panel itself to overlay, and as someone who has it in the TOP layer, i just find it aggravating when it pops up.

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -189,6 +189,11 @@
 		<default>42</default>
 		<min>0</min>
 	</option>
+	<option name="menu_force_show_popup" type="bool">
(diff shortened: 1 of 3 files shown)


trigg:
For reference, with it Always in Top it will never show up over a fullscreen video or game. with it Always in Overlay it never hides behind fullscreen video or games.

With this feature it only appears when requested

AKArien:
Well, this is precisely my problem. No objections for an option, i’d assume ?

PR #366, opened 2026-02-22
Update readme

AKArien:
A few readme changes i felt were necessary / would be welcome.

tell me what we want and what we don’t :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,22 @@ wf-shell is a repository which contains various components that can be used to b
 - wf-locker, a stylish and feature-full screen locker
 - wf-background, a simple background that can cycle through images
 
+## Configuration
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,30 +1,53 @@
...
 meson build --prefix=/usr --buildtype=release
-ninja -C build && sudo ninja -C build install
+ninja -C build
+sudo ninja -C build install
 ```
 
-# Configuration
+## Configuration
+
+The wf-shell components use a common config file located (by default) in `XDG_CONFIG_HOME/.config/wf-shell.ini`,
```
trigg:
```
    XDG_CONFIG_HOME
        Where user-specific configurations should be written (analogous to /etc).
        Should default to $HOME/.config.
```

Looks like you've accidentally doubled your .config directories :)

AKArien:
Also, are we looking to change the standard description of what wf-shell « is » ?
As in, old readme says « various components needed to built a fully functional DE », while the recent matrix chat room says « applications designed to be used as a Shell »
Not that they mean the opposite of each other, but i think i’d be better if everything referred to it the same way :)

PR #372, opened 2026-02-24
panel: window-list: Fix random crashes

soreau:
This fixes the case where some actions cause the panel to die, with the Gdk wayland error message "Error reading events from display: Invalid argument".


soreau:
As discussed in chat, this is the right thing to do anyway, so we might as well pull it, even if it doesn't fix the particular issue outlined in the commit message.

PR #376, opened 2026-02-27
panel : wp-mixer : allow different icon sizing in popups ; add vertical layout

AKArien:
Has a new css class for the part of the mixer that is shown in the popovers, `wireplumber-popup`. It is used to set the sizing of icons from a new option.

Add a setting to lay out the mixer with everything stacked vertically.


trigg (approved):
Concise & clean! Thank you again :)

PR #384, opened 2026-03-04
panl : wp-mixer : change css class names

AKArien:
Adjust css class names for the wp-mixer widget for consistency.
Give their own class to controls.

### DIFF
--- a/src/panel/panel.cpp
+++ b/src/panel/panel.cpp
@@ -538,7 +538,7 @@ void WayfirePanelApp::on_activate()
         {"panel/network_icon_size", ".network"},
         {"panel/volume_icon_size", ".volume"},
         {"panel/wp_icon_size", ".wp-mixer"},
-        {"panel/wp_popup_icon_size", ".wp-mixer-control"},
+        {"panel/wp_popup_icon_size", ".mute-toggle, .default-button"},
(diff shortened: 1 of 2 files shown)


src/panel/widgets/wp-mixer/wp-mixer.cpp
```diff
@@ -218,21 +218,21 @@ void WayfireWpMixer::init(Gtk::Box *container)
...
     button->add_css_class("flat");
     button->get_children()[0]->add_css_class("flat");
     button->set_child(main_image);
     button->show();
     sinks_box.add_css_class("outputs");
     sources_box.add_css_class("inputs");
     streams_box.add_css_class("streams");
-    out_in_wall.add_css_class("out-in");
-    in_streams_wall.add_css_class("in_streams");
+    out_in_wall.add_css_class("wall-out-in");
```
trigg: I wonder if the extra verbosity here is needed.

You can get to this element in a few ways already
`separator.out-in`
`separator.in_streams`

You can reference both with
`.wireplumber-popover separator`

do we need the separators to be more verbose?

AKArien: Ah right, didn't considier that

Then i'll just switch `in_streams` to `in-streams`

PR #391, opened 2026-03-07
locker: animation to indicate auth failed on password and fingerprint

trigg:


data/wf-locker
```diff
@@ -0,0 +1,2 @@
+auth    required pam_unix.so nodelay unlock_time=60
```
soreau, line 1:
```suggestion
auth    required pam_unix.so nodelay deny=3 unlock_time=10
```

data/wf-locker
```diff
@@ -0,0 +1,2 @@
...
+account required pam_unix.so nodelay unlock_time=60
```
soreau, line 2:
```suggestion
account required pam_unix.so nodelay deny=3 unlock_time=10
```

data/wf-locker
trigg: From my testing we never trigger the unlock lockout because we missed the deny=3 

But we do the lockout in our own software

Issue #325, opened 2026-01-18
Menu widget : removing an entire word with ctrl+del


AKArien:
As is, the menu doesn’t delete an entire word when pressing ctrl del. I’m not sure if this is tied to gtk or would need additional logic, but i would find it nice.

AKArien:
> By all means, if you're up for a challenge, give this a few pokes and see if you get a more favorable set of bugs :D

Hahahaha. Probably will at some point then, but can’t say it will be soon. we’ll see :)

AKArien:
Fixed by #395

PR #395, opened 2026-03-08
panel : menu : reduce search jank

AKArien:
Reduced jank on search, it now behaves like a proper text input, at the price of eating the first key when user attempts to type while it is not focused.

Addresses #325

### DIFF
--- a/src/panel/widgets/menu.cpp
+++ b/src/panel/widgets/menu.cpp
@@ -442,11 +442,9 @@ void WayfireMenu::load_menu_items_all()
 
 void WayfireMenu::on_search_changed()
 {
-    search_entry.set_text(search_contents);
-    search_entry.set_position(search_contents.length());
     if (menu_show_categories)
     {
-        if (search_contents.length() == 0)
+        if (search_entry.get_text().length() == 0)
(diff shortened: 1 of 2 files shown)


gtaborga:
I look forward to this change. I just discovered this problem when trying to use an on-screen keyboard.

AKArien:
Hey @trigg, in the end, since a few people seem to care about it being fixed asap, I would like it if we could get this current state reviewed and hopefully merged. My experiments for making it smoother probably won’t be ready any time soon.

AKArien:
@trigg hi, finally got to get around to it !
Took the liberty of merging your commit with mine, and i also added the check of the old system of verifying the gdk keyval string is equal to 1, as to avoid inserting strings for other keys (for instance, pressing shift adds « Shift » without this), as well as a comment explaining what is going on with it.

My impressions on this have not changed, while i’m not happy about this being needed and still want to overhaul the menu’s navigation, this is a huge step-up from prior behaviour. If it looks good to you, it looks good to me.

trigg (approved):
Looking good! Thanks for working on this

PR #396, opened 2026-03-09
xdpw: added a screencast chooser

trigg:
This PR adds a small GTK app to :
- Present user with a list of outputs & toplevels
- Allow user to select one
- Return this data to [xdpw ](https://github.com/emersion/xdg-desktop-portal-wlr)

Streaming toplevels require two extra protocols:
- ext-image-copy-capture-v1
- ext-foreign-toplevel-list-v1

Testing being done with:
- https://gitlab.gnome.org/-/snippets/19
- https://mozilla.github.io/webrtc-landing/gum_test.html

Packages & versions:

- sway 1.11
- xdpw 0.8.2-1
- wayfire (git latest)
- https://github.com/soreau/wf-copy-capture (git latest)

Quirks:
- stdout MUST NOT contain anything more than the prescribed reply. More likely we'll need to silence GTK warnings and errors or send them only to std:err. Cancel or failure to choose in any way MUST reply with empty stdout
- The chooser MUST detect when required protocols are missing for toplevel selection and present the user with the outputs instead.
- Currently, hotplug outputs is entirely theoretical. Patches welcome

TLDR:


trigg:
Part of me thinks matching the log-out menu from panel would be a more consistent look. 



But this could very easily be overkill to some users.

AKArien:
As discussed on the chatroom :

> Part of me thinks matching the log-out menu from panel would be a more consistent look

I would think it desirable too, not just for visual consistency but also for it to be "unmissable".

> But this could very easily be overkill to some users.

For the concern of an application spamming it, maybe we could try to do something like browsers do with alerts, and if we detect the same thing keeps trying and being denied, add a tickbox to automatically deny further requests ?

trigg:
Worth noting there is no mechanism to know who/what wishes to stream

soreau:
[portal-gum-test.webm](https://github.com/user-attachments/assets/7a8e9cc0-c454-459e-abcf-4e4b3ec73358)

PR #409, opened 2026-03-15
background: avoid setting the same image twice in a row

ammen99:
We recheck for new backgrounds every `background/cycle_timeout` seconds but in case nothing changed, we don't have to update anything.

In my case I have a very high res background image and had cycle_timeout=2, which pretty much caused a lot of CPU usage all the time.


trigg:
Thank you! Looks good

PR #411, opened 2026-03-21
Don't use menubuttons where we don't want buttons

trigg:
Refactor internal WayfireMenuButton to no longer actually be a menu button

MenuButton was originally chosen as it was the only way I had found to make a popover-menu for panel widgets. 

We put a fair bit of extra code to make MenuButton not actually be a menubutton. So this PR will standardise an interface for all Popup-based widgets without each one having to fight to not look like a button

Advantages over not having this PR:
- Fixed popover too big for entire screen at the source
- Moved autohide and layer shell logic out of individual widgets.
- Added option for popup-menu to follow mouse across other popup-menu widgets.
- class 'with-content' when a popup of any kind is attached
- class 'selected' when this specific widget has its popup shown
- users custom themes don't need to carefully unset all the button-specific css rules from their theme

### DIFF
--- a/data/css/default.css
+++ b/data/css/default.css
@@ -24,11 +24,21 @@
 .wf-panel .battery label,
 .wf-panel .battery:hover label,
 .wf-panel .window-button label,
+.wf-panel .window-button image,
 .wf-panel .window-button:hover label,
 .wf-panel .tray-button,
 .wf-panel .tray-button:hover,
 .wf-panel .notification button {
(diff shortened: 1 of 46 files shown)


trigg:
From my daily testing this is now in place.

BUT the change from button to box/image whatever internals has changed some of the extra padding added by button. The panel is more compact again, especially around menu button.

trigg:
fixes #418

PR #431, opened 2026-05-15
chore : clean up casts of WfOptions

AKArien:
As stated in chatrooms, the preffered way going forward is to let it be implicetely cast, and if impossible use .value() instead of explicitely casting

Maybe we should add a contributing section to the wiki that mentions it


soreau:
> As stated in chatrooms, the preffered way going forward is to let it be implicetely cast, and if impossible use .value() instead of explicitely casting

I do agree that this makes the code aesthetically cleaner, at least.

PR #432, opened 2026-05-15
chore : remove dead and commented-out code

AKArien:
The tray_smooth_scrolling_threshold was a leftover option from gtk3 version that is unused as of the gtk4 port.

Rest is a callaback getting set twice and commented-out code.


src/util/wf-autohide-window.cpp
```diff
@@ -607,17 +606,6 @@ void WayfireAutohidingWindow::set_active_popover(WayfireMenuButton& button)
 
     const bool should_grab_focus = this->active_button->is_keyboard_interactive();
 
-    /*
```
trigg: This particular part is being re-introduced in #430

PR #441, opened 2026-05-16
panel : mixer : fix styles for rename in fea9448

AKArien:


AKArien:
@trigg ok, this time, it *should* be good, but honestly at this point i’ve had so many failures in the past two days, would ask you to test to make sure lol

PR #442, opened 2026-05-17
panel: add config metadata for launchers

trigg:


mark-herbert42:
made the change in xml file but nothing happens on wcm - still no launchers section there. Does it need some more wcm/wf-config patches to work?

trigg:
I'm uncertain what would need to be done to add launchers to WCM, this was simply enough to stop a very large amount of warning spam on application startup

mark-herbert42:
I think I got the reason - in the command plugin the list of commands is declared as compound list option within the plugin code, but in the launchers code it is not so the launchers plugin simply reads the file and do some free-code parsing instead of using wf-config. That's  probably the reason why you can configure commands in wcm as dynamic list and add/remove/modify the commands while in launchers it does not show as configurable option even after the metadata xml is updated.

soreau:
> I think I got the reason - in the command plugin the list of commands is declared as compound list option within the plugin code, but in the launchers code it is not so the launchers plugin simply reads the file and do some free-code parsing instead of using wf-config. That's probably the reason why you can configure commands in wcm as dynamic list and add/remove/modify the commands while in launchers it does not show as configurable option even after the metadata xml is updated.

You are mostly correct. The thing is that for `autostart` and `command` plugins, wcm has custom code to support each. For wf-panel launchers, it does not have any special code. Ideally, wcm would support dynamic-list option type out-of-the-box, but it doesn't, because each case is slightly different. The good news is that we want to implement this, but we're not there yet.

PR #445, opened 2026-05-21
gtk3:Some MATE integragion and backporting bits from gtk4 version

mark-herbert42:
1. Left click on battery opens mate battery statistics dialog
2. Left click on volume open mate volume control dialog with advanced device selection, output/input profiles etc.
3. Backported language widget - now it shows keyboard language and allows to change with mouse click
4. Implemented panel unhide on keyboard language change and volume mute/unmute with keyboard shotcut
5. Backported configurable panel hide-unhide duration, 
6. Added notification center size config options.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "subprojects/wayland-logout"]
 	path = subprojects/wayland-logout
 	url = https://github.com/soreau/wayland-logout
+[submodule "subprojects/wf-json"]
+	path = subprojects/wf-json
+	url = https://github.com/WayfireWM/wf-json
--- a/meson.build
+++ b/meson.build
@@ -23,6 +23,7 @@ gtklayershell  = dependency('gtk-layer-shell-0', version: '>= 0.6', fallback: ['
 libpulse       = dependency('libpulse', required : get_option('pulse'))
 dbusmenu_gtk   = dependency('dbusmenu-gtk3-0.4')
 libgvc         = subproject('gvc', default_options: ['static=true'], required : get_option('pulse'))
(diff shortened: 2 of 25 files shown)


trigg:
This is a very mixed bag PR. 

Hardcoding in specific apps would be a reason to not take it, before even discussing if we intend to maintain a gtk3 fork alongside gtk4.

mark-herbert42:
> Hardcoding in specific apps 

Thanks,  fair point.  Updated to take the config option for cmmand to run on click instead of  hardcoding MATE utilities.

mark-herbert42:
> if we intend to maintain a gtk3 fork alongside gtk4.

I am 100% sure that it is not intended at all as gtk3 fork seems to be unchanged for a fairly long time - enough to understand it is abandoned. So my intention is not to keep it up to date with gk4 bleeding edge development which is too  hard, but keep a kind of usable gtk3 version that can be easily integrated with gtk3 based DE's like MATE and Xfce.  Those DE's will not in near future migrate to gtk4 (at least MATE which develops extremely slow this days).

ammen99:
@mark-herbert42 Just my 2 cents, but even if the panel is gtk4, you can use gtk3 apps from MATE with it. Also with CSS theming I expect that you can make the desktop look reasonably uniform across gtk4 and gtk3 applications.

AKArien:
Hi, unfortunately, maintaining a gtk3 branch in wf-shell isn’t planned.

We would invite you to fork and maintain your own version, and work with or without those desktops for backporting and enhancing functionality for this older version.

That being said, it seems you did have some original work (looking at the extra configuration for notifications), so feel free to propose these and more in a new PR !

PR #466, opened 2026-06-01
locker : make fingerprint updates count as activity

AKArien:
This causes the locker widgets to pop out when the fingerprint widget is updated, as it happens following user activity (fingerprint verification) or to notify the user of a change in sensor state


trigg:
As there are other signals, this triggers more often than it should. Ideally we need it to be inside the `verify-match` and `verify-no-match` branches, and certainly not outside of the `VerifyStatus` branch

AKArien:
Fair enough, i figured it would be best on anything, as to at least provide the user some sort of feedback that something is happening.

But looking back at it, i think it is best to move it to the `update` method, no ?

PR #468, opened 2026-06-01
autohide-window : add margins to the opposite edge if autohide is disabled

AKArien:
The margins being mostly a matter of style, i propose that if a margin is applied and autohide is disabled, space is reserved on both sides of the window.

Without this patch applied on wayfire master (thus upcoming 0.11, this is important because it used to have a bug) :
image

with :
image

### DIFF
--- a/metadata/dock.xml
+++ b/metadata/dock.xml
@@ -73,6 +73,11 @@
 		<_long>Space between the dock the edge</_long>
 		<default>0</default>
 	</option>
+	<option name="mirror_margin" type="bool">
+		<_short>Mirror margin</_short>
+		<_long>Apply the margin to both the edge and the windows.</_long>
+		<default>true</default>
+	</option>
 	<option name="edge_hotspot_size" type="int">
 		<_short>Edge hotspot size</_short>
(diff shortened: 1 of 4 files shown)


soreau:
It doesn't make sense to me, why you would want to workaround a bugged library version.

1) The bug should be fixed upstream.
2) This is not a good habit to make.
3) We can lock our gtk4-layer-shell version that is working as is.

AKArien:
>     1. The bug should be fixed upstream.

This is the intention, this behaviour is mainly aimed at the would-be correct behaviour, and happens to be less bad until it is
 
>     3. We can lock our gtk4-layer-shell version that is working as is.

The current version is « not working »

trigg:
So I think the reason I've not understood this is I couldn't work out which way around you were saying is wrong and which is right.

So looking at the setting, it pushes the panel out edge_margin pixels from the edge, which I believe is correct, but you're saying we should match the same margin on the far side too?

Sounds like call for an option to me, as I'd not expect margin to be applied on the inside as well as the outside, and dock could do with the same treatment potentially

AKArien:
@trigg better now ? :)

PR #469, opened 2026-06-02
Stream chooser live

soreau:
Implement live window previews. Requires latest wf-copy-capture plugin.


trigg:
Testing comes up great, code looks good!

soreau:
> Testing comes up great, code looks good!

@trigg Thanks for testing! I have updated from shm to dmabuf which changes many things. In theory, it should use less resources. Please let me know if it works well enough for you.

trigg:
LGTM

trigg:
There is a segfault on close, otherwise everything is good here. I'll be looking in to that later

soreau:
> There is a segfault on close, otherwise everything is good here. I'll be looking in to that later

It has already been fixed with the latest commit.

trigg:
Colours certainly look mixed on my case :D

PR #475, opened 2026-06-05
util: fix wifi AP password entry not always presenting itself

trigg:


trigg:
I should take this chance to name those variables better.

PR #481, opened 2026-06-09
fix not removing views from workspace preview issues

rpaciorek:
* fixed not removing closed or minimalised window from workspace-switcher grid preview
  * loop and line `WayfireWorkspaceWindow *w = (WayfireWorkspaceWindow*)widget;` caused removing fist processed overlay instead of related to a closed window
* fixed not removing closed window on not current workspace from workspace-switcher grid preview
  * `windows` was cleared in `clear_box` but `overlay` child was not removed which led to overlays being duplicated when switching workspaces


AKArien:
LGTM

PR #482, opened 2026-06-09
use boxflow in tray and launchers to allow multi rows/columns setup

rpaciorek:
* use FlowBox instead of Box in tray and launchers widgets to allow configure multi row panel
* fixed respecting `tray_icon_size` setting - add css class directly to icon


trigg:
initial look over, looks like a great idea. Do object to the misspelling of "horizontal"

trigg:
Ran and tested, good addition. Thank you!

PR #483, opened 2026-06-09
window-list: Switch live-previews to use ext-copy-capture instead of wlr-screencopy

soreau:


trigg:
Good work as always. Thanks again :)

PR #486, opened 2026-06-15
refresh command-output icons

mrsteve0924:
This PR adds support for dynamic command-output icons.

When the icon is specified as a filesystem path, it is reloaded after each command execution on the configured refresh interval. This allows icons for such such things as album art or weather icons to update automatically.

Icons specified by name (theme icons) continue to use the original static behavior and are not reloaded.

### DIFF
--- a/src/panel/widgets/command-output.cpp
+++ b/src/panel/widgets/command-output.cpp
@@ -14,7 +14,7 @@
 #include "command-output.hpp"
 
 static sigc::connection label_set_from_command(std::string command_line,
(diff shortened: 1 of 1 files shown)


trigg:
I'm labeling this as a bugfix as I think it's reasonable to assume this would work. Happy to merge once you pass uncrust ;)

mrsteve0924:
ok ill update the wiki to summarize the changes once merged. thanks

trigg:
Thank you :)

PR #490, opened 2026-06-20
panel : mixer : fix popup opening when doing mute_action

### LINKED ISSUE #488: `mute_quick_target` action also opening the mixer popup
## Summary

When `panel/mixer_middle_click_action` (or `..._right_click_action`) is set to
`mute_quick_target`, clicking that button also opens the mixer popup, same as
`show_mixer`. This makes the `mute_quick_target` action indistinguishable from
`show_mixer` in practice.

## Root cause

`mute_action` in `WayfireMixer::reload_config()` toggles the quick target's
mute state via:

```cpp
quick_target->button.set_active(!quick_target->button.get_active());
```

`quick_target->button` is a `Gtk::ToggleButton` whose `signal_toggled()` is
connected in `MixerControl::init()`:

```cpp
signals.push_back(mute_conn = button.signal_toggled().connect(
    [this, id] ()
{
    if (!parent->block_signals)
        parent->button->popup();   // <-- fires unconditionally on any toggle
    ignore = true;
    WpCommon::get().set_mute(id, button.get_active());
}));
```

Calling `set_active()` synchronously fires `signal_toggled()`, which calls
`parent->button->popup()` regardless of *why* the button was toggled. So the
quick-mute gesture ends up opening the popup as a side effect, even though
nothing asked it to.

## Proposed Fix

`MixerControl` already exposes `set_btn_status_no_callbk()` for updating the
toggle button's visual state without going through `signal_toggled()` (it's
used during initial setup for exactly this reason). `mute_action` now uses
that to update the button, and applies the mute itself via
`WpCommon::get().set_mute()`, bypassing the popup-opening side effect
entirely.

```cpp
auto mute_action = [&] (int c, double x, double y)
{
    if (!quick_target)
    {
        return;
    }

    guint32 id = wp_proxy_get_bound_id(WP_PROXY(quick_target->object));
    bool new_state = !quick_target->button.get_active();

    quick_target->set_btn_status_no_callbk(new_state);
    WpCommon::get().set_mute(id, new_state);
};
```

Only `src/panel/widgets/mixer/mixer.cpp` is touched; no header or
`mixer-control.cpp` changes needed.

## Testing

- Set:
  ```
  mixer_left_click_action   = show_mixer
  mixer_middle_click_action = mute_quick_target
  mixer_right_click_action  = show_quick_target
  ```
- Middle-click the panel widget: mute now toggles (icon updates) and the
  popup no longer opens.
- Confirmed muting via an external shortcut (not through this widget) still
  updates the icon correctly, since that path goes through
  `notify_is_muted_signal` independently of this lambda.
- Left/right click actions (`show_mixer`, `show_quick_target`) unaffected.

## Note for maintainers

`mute_action` (here) and `MixerControl`'s `signal_toggled` handler now both
call `WpCommon::get().set_mute(id, ...)` independently. This is intentional
to avoid touching `mixer-control.cpp`/`mixer.hpp`, but if the mute-setting
logic grows additional bookkeeping in the future, both call sites will need
to stay in sync.

AKArien:
Instead of activating the control’s button, which was opening the widget popup, manually set the state and prevent poping up.

Fixes #488

Co-authored by @froloffw7, please feel free to drop the email linked to your github account if you wish to be « properly » listed as a co-author.

### DIFF
--- a/src/panel/widgets/mixer/mixer.cpp
+++ b/src/panel/widgets/mixer/mixer.cpp
@@ -98,7 +98,15 @@ void WayfireMixer::reload_config()
             return; // no quick_target means we have nothing to change by clicking
         }
 
-        quick_target->button.set_active(!quick_target->button.get_active());
+        guint32 id = wp_proxy_get_bound_id(WP_PROXY(quick_target->object));
+        bool mute  = !quick_target->button.get_active();
+
+        // pretend the mixer control is the source of the change to prevent a popup
+        objects_to_controls[quick_target->object]->ignore = true;
+        // still set visual state
+        objects_to_controls[quick_target->object]->set_btn_status_no_callbk(mute);
+        quick_target->set_btn_status_no_callbk(mute);
+        WpCommon::get().set_mute(id, mute);
     };
 
     if (str_wp_left_click_action.value() == "show_mixer")


trigg:
Looks good, thank you

PR #493, opened 2026-06-22
panel: menu launchers double fork to not remain child of panel

trigg:
Anything launched via menu currently dies if the panel dies.

This will double-fork and close the intermediate child instantly. This causes the end-child to be picked up by PID1 and to outlive a crashing or closing panel.


trigg:
Didn't fix it under enough circumstances, and is not at all a nice approach to this.

PR #505, opened 2026-07-30
panel: fix occasional crash

trigg:
Bug report and fix in one, sorry

To reproduce crash:

get a program with an offending popup. I used xfce4-terminal's multi-line paste warning but also good were open/save as dialogs and almost all of android studio.

- Start wf-panel
- Add monitor
- Remove monitor
- cause popup

This would always cause a crash, but work fine if no hotplugging happened prior

The cause was leaked listeners continuing after the widget had deconstructed.


trigg:
As it's simple an unobtrusive enough, and has fixed all my current crashes, I'm merging this now

PR #507, opened 2026-07-31
panel: split menu widget files up

trigg:
The contents of menu.Xpp had become too large and unwieldy.

So we have split them into class files.

More cleaning could be done, but this was a clear first step.

Not for before release, as it is not actively changing anything.


trigg:
So I have changed my mind, this is getting pulled in before release because the menu-speed PR #508 uses it, and is a very good idea to bring in before hand

PR #510, opened 2026-07-31
panel : default menu widgets to be centered horizontally and vertically

AKArien:
This is at the moment the only kind of alignment that is wanted, and will probably remain so forever. Set the alignment once in wf-popover and let widgets handle themselves if they ever want somethig else, as well as their container box.

This fixes an issue with certain widgets not being horizontally centered when the panel is in a vertical layout (positioned left or right)

### DIFF
--- a/src/panel/panel.cpp
+++ b/src/panel/panel.cpp
@@ -106,26 +106,25 @@ class WayfirePanel::impl
             right_box.measure(orientation, -1, rmin, rnat, rminb, rnatb);
             center_box.measure(orientation, -1, mmin, mnat, mminb, mnatb);
 
+            center_box.set_halign(Gtk::Align::CENTER);
+            center_box.set_valign(Gtk::Align::CENTER);
+
             if (is_horizontal)
             {
                 content_box.set_size_request((std::max(lnat, rnat) * 2) + mnat, -1);
 
                 left_box.set_halign(Gtk::Align::END);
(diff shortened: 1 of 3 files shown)


trigg:
Looking good for the most part! One issue on window-list in horizontal from my tests, I'll dig around that now

trigg:
So you've got a lot of `container` set_halign/set_valign, and that's the parent (left/center/right) and has its own setting prior :) Should only be setting in the widget itself.

AKArien:
> So you've got a lot of `container` set_halign/set_valign, and that's the parent (left/center/right) and has its own setting prior :) Should only be setting in the widget itself.

hm, right, i could have sworn that it didn’t look quite right if i didn’t do that, but… well it sure does, so i’m not sure what that was about… i need to revise my wf-panel widgets tree :sweat_smile: 

I’ve pushed the original-corrected version, hopefully it’s good now ?

trigg:
My counter proposal, make our button do the heavy lifting.

Only widget with an issue that I saw was window-list, but thats not to do with this PR.

AKArien:
> My counter proposal, make our button do the heavy lifting.
> 
> Only widget with an issue that I saw was window-list, but thats not to do with this PR.

Hm, sorry, i really don’t think i understand what you mean at all ?

trigg:
Taken #516 which included this contents. Thanks!

PR #526, opened 2026-08-03
Use more common icon name for wf-locker-pin

City-busz:
The `lock-symbolic` cannot be found in the Adwaita theme. Use `system-lock-screen-symbolic` instead.

### DIFF
--- a/data/wf-locker-pin.desktop
+++ b/data/wf-locker-pin.desktop
@@ -2,7 +2,7 @@
 Version=1.0
 Type=Application
 Exec=wf-locker-pin
-Icon=lock-symbolic
+Icon=system-lock-screen
 Terminal=false
 Name=Wf Shell Change PIN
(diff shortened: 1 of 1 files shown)


AKArien:
while i'm on board, the [standard icons spec](https://specifications.freedesktop.org/icon-naming/latest/) only has `system-lock-screen`, not symbolic, and as such, through rapid, manual looking through icon themes, it seems to me it's about as common as the the current `lock-symbolic`.

I believe it would also simply be more appropriate to not use a `-symbolic` icon for an application, so i rather propose simply using `system-lock-screen`.

City-busz:
Updated to use the non-symbolic version.

trigg:
Yeah, fair enough. I chose a quick easy icon from what I had and didn't think a second time about it, so that's my mistake.

Thanks for the patch!

PR #531, opened 2026-08-07
panel : standalone power widget

AKArien:
Move the functionality to set power profiles to it’s own widget.

In my opinion, it makes more sense for it to be so than for it to be tacked on to the battery, as the battery and power profile mode are only loosely related. This also makes the icon showing the power mode a more « proper » feature than it being a fallback icon for the `battery` widget if the device doesn’t have a battery.

### DIFF
--- a/metadata/panel.xml
+++ b/metadata/panel.xml
@@ -235,6 +235,18 @@ If full_span is off, both sides of the panel will take the same amount of space,
 	</option>
 	</group>
 	<group>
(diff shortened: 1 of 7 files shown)


r0man1an:
Would be nice to have the option to chose whether it is in the battery applet or it's own. I personally think its better to keep it in the battery applet

trigg:
I'd still find this a problem as I'd automatically go to battery to change my power profile

PR #533, opened 2026-08-09
Update and clean up ci

AKArien:
Doesn’t change how it actually runs, but is a lot prettier imo

Also bumped the checkout actions from v2 to latest (v7)


trigg:
Neat!

PR #539, opened 2026-08-09
panel: menu: fix rightclick response time

trigg:


trigg:
waving this through as it's very simple

PR #540, opened 2026-08-09
data: fix meson.build

hustlerone:
hardcoded paths are le bad


trigg:
In the case of XDPW I think this will work, but for PAM this would only break it.

We would need it to be in /etc/pam.d or /usr/lib/pam.d but not at the whims of prefix locations.

hustlerone:
So, what do we do then? This is what Meson says:
https://mesonbuild.com/Builtin-options.html

and normally it would be /etc/pam.d/

trigg:
Fair play, that's on me, I misread it as another variable. Well it's worth a look at. Do we know any use-cases we can test where sysconfdir is set to something else and PAM sees it the same way?

hustlerone:
You can override where sysconfdir points to when running iirc. By default it's /etc/ or whatever the distro chooses for sysconf directory to be

trigg:
Yes, the more I read into it the more I'm sure you're correct.

Thanks for being patient with me!

PR #542, opened 2026-08-09
panel : mixer : use microphone icons for input devices

AKArien:


trigg:
Solid idea. I hadn't thought about that but it would certainly help!

PR #543, opened 2026-08-09
util: added '-v' and '--version' 

trigg:
creates a version constant string on compile which is made up of
- version in meson 
- branch name from git
- short commit code from git

the hope is to make issues more helpful by including a version. Some warning should be due that packaged builds are not likely to include either, and I suspect that will be fine.

util/wf-shell-app.cpp handles this in case '-v' or '--version' are CLI args


trigg:
TODO:
- locker
- locker-pin
- stream-chooser

PR #547, opened 2026-08-11
ci: add GCC and Clang build matrix

### LINKED ISSUE #534: ci : use matrix testing for build
As seen on #530, the ci (and most users) uses GCC, it would be a good idea to test for clang as well. Could also test with glibc along with musl.

With #511, i am also wondering if we should have a list of supported versions that are tested for ? but that would probably be way, way too overkill

w3lld1:
## Summary

- I converted the Alpine build job into a compiler matrix covering GCC and Clang.
- I pass the selected `CC` and `CXX` explicitly to Meson for each matrix entry.
- I kept both builds on the existing musl/libstdc++/BFD environment so compiler differences are isolated.

## Validation

- `actionlint .github/workflows/ci.yaml`
- YAML matrix assertions for both compiler entries and their build environment
- Alpine package/index check for `g++`, `clang21`, `clang-21`, and `clang++-21`
- `git diff --check upstream/master...HEAD`

I deferred the complete container builds to GitHub Actions because the local Docker socket is unavailable.

Fixes #534


trigg:
Looks good, thank you !

PR #551, opened 2026-08-25
panel: fix ipc error detection

muchnameless:
workspace switcher widget window previews disappear when a window contains the string "error" in the title

this pattern existed in multiple places / widgets


AKArien:
Nice catch ! LGTM

trigg:
Thank you!

PR #559, opened 2026-09-12
panel: remove unused pulseaudio include from brightness

cjmayo:
Compiles fine for me. Can't find any pa_/PA_ in brightness.

Allows building wf-shell without `volume-widget` on systems lacking pulseaudio.


trigg:
Good catch, thanks for the clean up!

PR #560, opened 2026-09-18
panel: add options to prefer symbolic icons

mrsteve0924:
This is a draft PR to get feedback on the approach before expanding symbolic icon support to additional panel widgets.

This adds two new panel options:
`prefer_symbolic_widget_icons = false`
`prefer_symbolic_tray_icons = false`

`prefer_symbolic_widget_icons` currently applies to the volume widget. When enabled, the volume widget prefers the `-symbolic` variants of its icons, with the existing icons used as fallbacks.

`prefer_symbolic_tray_icons` applies to StatusNotifier tray items. When enabled, the tray first attempts to load `<IconName>-symbolic`, then falls back to the application's original `IconName`, and finally to the existing pixmap fallback.

Both options default to `false`, so the existing behavior is unchanged unless explicitly enabled.

I separated tray icons from panel widget icons because tray icons are provided by external applications, while widget icons are selected by wf-shell.

If this general approach makes sense, I would like to expand `prefer_symbolic_widget_icons` to other appropriate panel widgets in follow-up changes.

Assisted-by: ChatGPT:GPT-5.6 Sol
Signed-off-by: steve finkel <mrsteve0924@verizon.net>


trigg:
Clear improvement for bluetooth icon, but no improvement for steam.

Novel idea, and I am pleased with the results so far.

mrsteve0924:
what is the difference between those 2 lines in the image?

mrsteve0924:
-any reason to not expand this to cover the other current panel widgets? 
-should i create a new group for these symbolic icon preferences in the xml or just try to wedge them in somewhere?

AKArien:
imo this is great !

If i may suggest : to me, a toggle for the whole panel should be fine.
Applicable icon choices would go through a new utility in src/util/icon-select.cpp, which does the check and switches the icon to the prefered version.
It could also then fall back to non-symbolic icons if it's found they aren't available and the user prefers symbolic, and the opposite.

mrsteve0924:
@AKArien I like this ides, "Applicable icon choices would go through a new utility in src/util/icon-select.cpp, which does the check and switches the icon to the prefered version."

thanks

PR #561, opened 2026-09-21
window-list: fix panel crash when child toplevel changes parent

mrsteve0924:
Panel will crash when clicking on an open app in the window-list.
Mostly happens when opening a file from thunar then clicking on the  thunar window-list widget.

Remove a child from its previous parent when its parent changes. 
Also avoid dereferencing stale child entries when handling window-list clicks.


mrsteve0924:
the crash is difficult to reproduce.  i was able to crash the panel again, but using different click patterns on the window-list widget. here is the backtrace (link below).

https://dpaste.com//EL6JGPUBE#wrap

